PR #3, opened 2015-10-07
SCMNavigator

jglick:
- [X] update baseline
- [x] add `SCMNavigator`
- [x] introduce `SCMSourceObserver`
- [x] perform arbitrary reformattings

@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMFileSystem.java
+++ b/src/main/java/jenkins/scm/api/SCMFileSystem.java
@@ -102,7 +102,11 @@ public abstract class SCMFileSystem {
     public static SCMFileSystem of(@NonNull SCM scm, @CheckForNull SCMRevision rev) {
         scm.getClass(); // throw NPE if null
         SCMFileSystem fallBack = null;
-        for (Builder b : Jenkins.getInstance().getExtensionList(Builder.class)) {
+        Jenkins j = Jenkins.getInstance();
+        if (j == null) {
+            return fallBack;
+        }
+        for (Builder b : j.getExtensionList(Builder.class)) { // TODO 1.572+ ExtensionList.lookup
             SCMFileSystem inspector = b.build(scm, rev);
             if (inspector != null) {
                 if (inspector.isFixedRevision()) {
@@ -144,7 +148,11 @@ public abstract class SCMFileSystem {
                                    @CheckForNull SCMRevision rev) {
         source.getClass(); // throw NPE if null
         SCMFileSystem fallBack = null;
-        for (Builder b : Jenkins.getInstance().getExtensionList(Builder.class)) {
+        Jenkins j = Jenkins.getInstance();
+        if (j == null) {
+            return fallBack;
+        }
+        for (Builder b : j.getExtensionList(Builder.class)) { // TODO 1.572+ ExtensionList.lookup
             SCMFileSystem inspector = b.build(source, head, rev);
             if (inspector != null) {
                 if (inspector.isFixedRevision()) {
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package jenkins.scm.api;
+
+import hudson.ExtensionPoint;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.TaskListener;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+/**
+ * Defines how we can scan for the presence of repositories.
+ */
+public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator> implements ExtensionPoint {
+
+    protected SCMNavigator() {}
+
+    /**
+     * Looks for SCM sources in a configured place.
+     * @param context who is asking
+     * @param listener a listener to be notified of progress
+     * @return a map from project names, to a list of potential SCM sources as in {@code MultiBranchProject.getSCMSources}; do not call {@link SCMSource#setOwner} on them
+     * @throws IOException if scanning fails
+     * @throws InterruptedException if scanning is interrupted
+     */
+    public abstract @Nonnull Map<String,? extends List<? extends SCMSource>> discoverSources(@Nonnull SCMSourceOwner context, @Nonnull TaskListener listener) throws IOException, InterruptedException;
+
+    // TODO this probably needs some way of customizing projects that get created, for example to add a link to GitHub in the description
+
+    @Override public SCMNavigatorDescriptor getDescriptor() {
+        return (SCMNavigatorDescriptor) super.getDescriptor();
+    }
+
+}
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -48,7 +48,8 @@ public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator>
      */
     public abstract void visitSources(@Nonnull SCMSourceObserver observer) throws IOException, InterruptedException;
 
-    @Override public SCMNavigatorDescriptor getDescriptor() {
+    @Override
+    public SCMNavigatorDescriptor getDescriptor() {
         return (SCMNavigatorDescriptor) super.getDescriptor();
     }
 
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -33,7 +33,9 @@ import java.util.Map;
 import javax.annotation.Nonnull;
 
 /**
- * Defines how we can scan for the presence of repositories.
+ * An API for discovering new and navigating already discovered {@link SCMSource}s within an organization.
+ * An implementation does not need to cache existing discoveries, but some form of caching is strongly recommended
+ * where the backing provider of repositories has a rate limiter on API calls.
  */
 public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator> implements ExtensionPoint {
 
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -36,6 +36,7 @@ import javax.annotation.Nonnull;
  * An API for discovering new and navigating already discovered {@link SCMSource}s within an organization.
  * An implementation does not need to cache existing discoveries, but some form of caching is strongly recommended
  * where the backing provider of repositories has a rate limiter on API calls.
+ * @since FIXME
  */
 public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator> implements ExtensionPoint {
 
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -26,10 +26,7 @@ package jenkins.scm.api;
 
 import hudson.ExtensionPoint;
 import hudson.model.AbstractDescribableImpl;
-import hudson.model.TaskListener;
 import java.io.IOException;
-import java.util.List;
-import java.util.Map;
 import javax.annotation.Nonnull;
 
 /**
@@ -44,15 +41,12 @@ public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator>
 
     /**
      * Looks for SCM sources in a configured place.
-     * @param context who is asking
-     * @param listener a listener to be notified of progress
-     * @return a map from project names, to a list of potential SCM sources as in {@code MultiBranchProject.getSCMSources}; do not call {@link SCMSource#setOwner} on them
+     * After this method completes, no further calls may be made to the {@code observer} or its child callbacks.
+     * @param observer a recipient of progress notifications and a source of contextual information
      * @throws IOException if scanning fails
      * @throws InterruptedException if scanning is interrupted
      */
-    public abstract @Nonnull Map<String,? extends List<? extends SCMSource>> discoverSources(@Nonnull SCMSourceOwner context, @Nonnull TaskListener listener) throws IOException, InterruptedException;
-
-    // TODO this probably needs some way of customizing projects that get created, for example to add a link to GitHub in the description
+    public abstract void visitSources(@Nonnull SCMSourceObserver observer) throws IOException, InterruptedException;
 
     @Override public SCMNavigatorDescriptor getDescriptor() {
         return (SCMNavigatorDescriptor) super.getDescriptor();
--- a/src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
@@ -28,6 +28,10 @@ import hudson.model.Descriptor;
 import hudson.model.TopLevelItemDescriptor;
 import javax.annotation.CheckForNull;
 
+/**
+ * Definition of a kind of {@link SCMNavigator}.
+ * @since FIXME
+ */
 public abstract class SCMNavigatorDescriptor extends Descriptor<SCMNavigator> {
 
     protected SCMNavigatorDescriptor() {}
--- a/src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
@@ -1,0 +1,48 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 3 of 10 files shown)


src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -102,7 +102,11 @@ public static SCMFileSystem of(@NonNull SCM scm) {
     public static SCMFileSystem of(@NonNull SCM scm, @CheckForNull SCMRevision rev) {
         scm.getClass(); // throw NPE if null
         SCMFileSystem fallBack = null;
-        for (Builder b : Jenkins.getInstance().getExtensionList(Builder.class)) {
+        Jenkins j = Jenkins.getInstance();
+        if (j == null) {
+            return fallBack;
```
tfennelly: Why not just explicitly return `null`?

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -144,7 +148,11 @@ public static SCMFileSystem of(@NonNull SCMSource source, @NonNull SCMHead head,
                                    @CheckForNull SCMRevision rev) {
         source.getClass(); // throw NPE if null
         SCMFileSystem fallBack = null;
-        for (Builder b : Jenkins.getInstance().getExtensionList(Builder.class)) {
+        Jenkins j = Jenkins.getInstance();
+        if (j == null) {
+            return fallBack;
```
tfennelly: and here

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -0,0 +1,58 @@
...
+import hudson.model.TaskListener;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+/**
+ * Defines how we can scan for the presence of repositories.
+ */
+public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator> implements ExtensionPoint {
```
tfennelly: `SCMScanner` might be a better name perhaps (if it's not already gone)? "Navigator" suggests (to me at least) that there's a tree/graph of some sort that's being traversed/navigated, which I don't think is the situation.

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Looks for SCM sources in a configured place.
+     * @param context who is asking
+     * @param listener a listener to be notified of progress
+     * @return a map from project names, to a list of potential SCM sources as in {@code MultiBranchProject.getSCMSources}; do not call {@link SCMSource#setOwner} on them
+     * @throws IOException if scanning fails
+     * @throws InterruptedException if scanning is interrupted
+     */
+    public abstract @Nonnull Map<String,? extends List<? extends SCMSource>> discoverSources(@Nonnull SCMSourceOwner context, @Nonnull TaskListener listener) throws IOException, InterruptedException;
```
tfennelly: Would it make sense to have a `ProjectSourceSet` type to encapsulate `Map<String,? extends List<? extends SCMSource>>` and the operations you might want to perform on that? Not sure about others, but I go bug-eyed looking at things like that.

src/main/java/jenkins/scm/impl/SingleSCMNavigator.java
```diff
@@ -0,0 +1,76 @@
...
+            return "Single Repository";
+        }
+
+        @Override public SCMNavigator newInstance(String name) {
+            return null;
+        }
+
+    }
+
+}
```
tfennelly: Once could guess what this class is for, but maybe some javadocs would be good.

src/main/java/jenkins/scm/api/SCMNavigator.java
amuniz: `SCMSourceScanner` was my proposal :smile:

src/main/java/jenkins/scm/api/SCMNavigator.java
amuniz: +1
IMO `SCMSourceSets` would be a better name.

amuniz:
:bee:

src/main/java/jenkins/scm/api/SCMFileSystem.java
jglick: This is just temporary code until I can bump up to 1.572 or newer and simplify. Maybe I should just do that now in `mercurial` and `subversion`; there is not much point in retaining compatibility with 1.568–1.579.

src/main/java/jenkins/scm/api/SCMFileSystem.java
jglick: Ditto, this was just to shut up FindBugs and can be simplified further with a new core.

src/main/java/jenkins/scm/api/SCMNavigator.java
jglick: There *is* typically a tree that is being traversed, in the case of an organization. Anyway I have no strong opinion but work it out with @stephenc. (BTW renaming it now is going to be more work than it would have been yesterday.)

jglick: `SCMSourceSoláthróir`? :-)

src/main/java/jenkins/scm/api/SCMNavigator.java
jglick: Perhaps. The method also will need the ability to send the `context` some metadata attributes such as icon, so I suspect this will need to be deprecated and replaced soon anyway.

src/main/java/jenkins/scm/api/SCMNavigator.java
tfennelly: @jglick ah go an-mhaith :)

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -0,0 +1,58 @@
...
+import hudson.ExtensionPoint;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.TaskListener;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+/**
+ * Defines how we can scan for the presence of repositories.
```
stephenc:
```
/**
 * An API for discovering new and navigating already discovered {@link SCMSource}s within an organization.
 * An implementation does not need to cache existing discoveries, but some form of caching is strongly recommended
 * where the backing provider of repositories has a rate limiter on API calls.
 * @since FIXME
 */
```

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: `SCMSurveyor` is perhaps an alternative if people really object to Navigator

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: I think it could return a `SCMAtlas` from an `SCMAtlasRequest`

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator> implements ExtensionPoint {
+
+    protected SCMNavigator() {}
+
+    /**
+     * Looks for SCM sources in a configured place.
```
stephenc: Clarify whether this is a live map (i.e. where traversing the map may be blocking) or a snapshot map (i.e. where getting the map may be blocking), and what the expected cost is to the caller.

src/main/java/jenkins/scm/api/SCMSourceDescriptor.java
```diff
@@ -128,7 +128,11 @@ public boolean isUserInstantiable() {
     public static List<SCMSourceDescriptor> forOwner(Class<? extends SCMSourceOwner> clazz,
                                                      boolean onlyUserInstantiable) {
         List<SCMSourceDescriptor> result = new ArrayList<SCMSourceDescriptor>();
-        for (Descriptor<SCMSource> d : Jenkins.getInstance().getDescriptorList(SCMSource.class)) {
+        Jenkins j = Jenkins.getInstance();
```
stephenc: // TODO Jenkins 1.572+

src/main/java/jenkins/scm/api/SCMSourceOwners.java
```diff
@@ -66,7 +67,11 @@
          * {@inheritDoc}
          */
         public Iterator<SCMSourceOwner> iterator() {
-            return Jenkins.getInstance().getAllItems(SCMSourceOwner.class).iterator();
+            Jenkins j = Jenkins.getInstance();
```
stephenc: // TODO

src/main/java/jenkins/scm/api/SCMSourceOwners.java
```diff
@@ -78,7 +83,11 @@
          * {@inheritDoc}
          */
         public Iterator<SCMSourceOwner> iterator() {
-            return new IteratorImpl(Jenkins.getInstance().getExtensionList(Enumerator.class));
+            Jenkins j = Jenkins.getInstance();
```
stephenc: // TODO

src/main/java/jenkins/scm/api/SCMNavigator.java
jglick: Another option (CC @stephenc as usual, plus @recena in this case) would be to introduce

```java
public abstract class SCMSourceObserver {
    public abstract SCMSourceOwner getContext();
    public abstract TaskListener getListener();
    public abstract ProjectObserver observe(String projectName);
    public abstract void addAttribute(String key, Object value);
    public static abstract class ProjectObserver {
        public abstract void addSource(SCMSource source);
        public abstract void addAttribute(String key, Object value);
    }
}
```

and then have `SCMNavigator` (or whatever) just have something like

```java
public abstract void discoverSources(SCMSourceObserver observer) throws IOException, InterruptedException;
```

The main advantage is ease of introducing compatible API extensions.

stephenc:
:bug: code style does not match https://github.com/jenkinsci/scm-api-plugin/blob/master/CONTRIBUTING.md

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: I'm liking that, but the method should perhaps be `visitSources` and taking a `SCMSourceVisitor`

src/main/java/jenkins/scm/api/SCMSourceDescriptor.java
jglick: There is already such a TODO comment a few lines down.

src/main/java/jenkins/scm/api/SCMSourceOwners.java
jglick: What is to be done?

src/main/java/jenkins/scm/api/SCMSourceOwners.java
jglick: Again there is already a comment a few lines down.

src/main/java/jenkins/scm/api/SCMNavigator.java
jglick: “Visit” is fine with me in general but I was trying to retain some consistency with `SCMHeadObserver`.

Would you pick a different name for `SCMNavigator` in light of this discussion?

recena: +1 @jglick, it seems clear.

src/main/java/jenkins/scm/impl/SingleSCMNavigator.java
jglick: Yes I can add some.

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: I am fine keeping `SCMSourceObserver` as I had forgotten `SCMHeadObserver`... I still think the method name should use `visit`

src/main/java/jenkins/scm/api/SCMNavigator.java
jglick: Would be irrelevant in the case of a visitor. (Then you just need to emphasize that it is illegal to call visitor methods after the visitation has concluded.)

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: A you use a good navigator to visit many locations efficiently... I think we are calling it `SCMNavigator`

src/main/java/jenkins/scm/impl/SingleSCMNavigator.java
jglick: (resolved)

amuniz:
:bee: (again)

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: ok

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Creates a default navigator, if there should be one displayed.
+     * May be interpreted by {@code jenkins.branch.SpecificDescriptor}.
+     * If returning non-null, you may also provide a {@code newInstanceDetail} view as per {@link TopLevelItemDescriptor}.
+     * @param name a name supplied by the user which may be used as a hint for preconfiguration
+     * @return a navigator with a default configuration, or null if it should not be advertised this way
+     */
+    public abstract @CheckForNull SCMNavigator newInstance(@CheckForNull String name);
```
stephenc: style guide

src/main/java/jenkins/scm/api/SCMSourceDescriptor.java
stephenc: ok

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+ * Callback used by {@link SCMNavigator}.
+ * @since FIXME
+ */
+public abstract class SCMSourceObserver {
+
+    /**
+     * Indicates who is asking for sources.
+     * @return a contextual item, typically a {@code OrganizationFolder}
+     */
+    public abstract @Nonnull SCMSourceOwner getContext();
```
stephenc: style guide

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Provides a way of reporting progress.
+     * @return a logger
+     */
+    public abstract @Nonnull TaskListener getListener();
```
stephenc: style guide

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Declare that a new “project” such as a source repository has been found.
+     * @param projectName a name of the project, such as a repository name within an organization; may be used as an {@link Item#getName}
+     * @return a secondary callback to customize the project, on which you must call {@link ProjectObserver#complete}
+     * @throws IllegalArgumentException if this {@code projectName} has already been encountered
+     */
+    public abstract @Nonnull ProjectObserver observe(@Nonnull String projectName) throws IllegalArgumentException;
```
stephenc: style guide

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Adds extra metadata about the overall organization.
+     * Currently no metadata keys are defined; placeholder for description, icon, URL, etc.
+     * @param key a predefined attribute name
+     * @param value some value, of a type defined by the attribute
+     * @throws IllegalArgumentException if the attribute name is unrecognized
+     * @throws ClassCastException if the attribute value is inappropriate for its type
+     */
+    public abstract void addAttribute(@Nonnull String key, @Nullable Object value) throws IllegalArgumentException, ClassCastException;
```
stephenc: Should this not be `CheckForNull`?

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return a secondary callback to customize the project, on which you must call {@link ProjectObserver#complete}
+     * @throws IllegalArgumentException if this {@code projectName} has already been encountered
+     */
+    public abstract @Nonnull ProjectObserver observe(@Nonnull String projectName) throws IllegalArgumentException;
+
+    /**
+     * Adds extra metadata about the overall organization.
+     * Currently no metadata keys are defined; placeholder for description, icon, URL, etc.
+     * @param key a predefined attribute name
+     * @param value some value, of a type defined by the attribute
```
stephenc: Does a `{@code null}` value remove the existing attributes or is it a valid value.

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Declare that a new “project” such as a source repository has been found.
+     * @param projectName a name of the project, such as a repository name within an organization; may be used as an {@link Item#getName}
+     * @return a secondary callback to customize the project, on which you must call {@link ProjectObserver#complete}
+     * @throws IllegalArgumentException if this {@code projectName} has already been encountered
+     */
+    public abstract @Nonnull ProjectObserver observe(@Nonnull String projectName) throws IllegalArgumentException;
+
+    /**
+     * Adds extra metadata about the overall organization.
```
stephenc: What happens if you add the same key with different values? is it last wins or will there be multiple values? Can we remove attributes?

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -0,0 +1,98 @@
...
+
+        /**
+         * Adds extra metadata about a specific project.
+         * Currently no metadata keys are defined; placeholder for description, icon, URL, etc.
+         * @param key a predefined attribute name
+         * @param value some value, of a type defined by the attribute
+         * @throws IllegalArgumentException if the attribute name is unrecognized
+         * @throws ClassCastException if the attribute value is inappropriate for its type
+         */
+        public abstract void addAttribute(@Nonnull String key, @Nullable Object value) throws IllegalArgumentException, ClassCastException;
```
stephenc: Same questions w.r.t. outer class's `addAttribute`

src/main/java/jenkins/scm/api/SCMSourceOwners.java
stephenc: ok

src/main/java/jenkins/scm/api/SCMSourceOwners.java
stephenc: I think it would be fine to have Jenkins 1.590+ use `getActiveInstance()`

src/main/java/jenkins/scm/impl/SingleSCMNavigator.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Extension
+    public static class DescriptorImpl extends SCMNavigatorDescriptor {
+
+        @Override
+        public String getDisplayName() {
+            return "Single Repository";
+        }
+
+        @Override public SCMNavigator newInstance(String name) {
```
stephenc: style guide

src/main/resources/jenkins/scm/impl/SingleSCMNavigator/config.jelly
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stephenc: MIT license please (Have been trying to keep this present for all .jelly files)

stephenc:
Technically almost there, needs some clarification in parts of the javadoc

src/main/java/jenkins/scm/api/SCMSourceObserver.java
jglick: Well, it depends on the `key`, so this seemed like one the grey areas that `@Nullable` was intended for. No strong opinion.

stephenc:
:bee:

src/main/resources/jenkins/scm/impl/SingleSCMNavigator/config.jelly
jglick: (resolved)

PR #6, opened 2016-01-29
[JENKINS-32493] Adapt to new parent POM.

andresrc:
* [x] Adapt POM to new plugin parent POM.
* [x] Adapt Jelly to pass injected tests.
* [x] Adapt Javadoc to build under Java 8.
* [x] Some stream closing clean up.
* [x] Remove TODOs related to using `ExtensionList.lookup` where applicable.

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMSourceOwners.java
+++ b/src/main/java/jenkins/scm/api/SCMSourceOwners.java
@@ -23,6 +23,7 @@
  */
 package jenkins.scm.api;
 
+import com.google.common.collect.Iterators;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.ExtensionPoint;
@@ -69,7 +70,7 @@ public abstract class SCMSourceOwners implements ExtensionPoint, Iterable<SCMSou
         public Iterator<SCMSourceOwner> iterator() {
             Jenkins j = Jenkins.getInstance(); // TODO 1.590+ getActiveInstance
             if (j == null) {
-                return Collections.emptyIterator();
+                return Iterators.emptyIterator();
             }
             return j.getAllItems(SCMSourceOwner.class).iterator();
         }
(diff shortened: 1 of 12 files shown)


andresrc:
@reviewbybees

src/main/java/jenkins/scm/api/SCMSourceOwners.java
```diff
@@ -69,7 +70,7 @@
         public Iterator<SCMSourceOwner> iterator() {
             Jenkins j = Jenkins.getInstance(); // TODO 1.590+ getActiveInstance
```
amuniz: Since you are cleaning up things, you could clean this one too (core is bumped to 1.596.1 in this PR).

andresrc: In this case, using `getActiveInstance` changes behavior: exception instead of graceful fallback with an empty iterator. I'm cleaning up other TODOs that don't change bechavior.

amuniz:
:bee:

andresrc:
@reviewbybees done

stephenc:
:+1: as maintainer

src/main/java/jenkins/scm/api/SCMSourceOwners.java
jglick: Regardless of the final behavior you select, leaving in a TODO comment referring to a Jenkins version you are now past is a clear mistake. If you intend to retain the null check, then the comment should be deleted.

stephenc: @andresrc 
Answer the following questions please:

* Is this code running in a plugin?
* Is this code only ever running in the same JVM as the Jenkins singleton?

If the answer to the above two questions are both yes (and let's be fair here, the first one has to be a yes) then `Jenkins.getInstance()` will **never** be `null` (or if it is `null` then that is a bug in Jenkins core)

So either 
* this is a class that is not intended to be called from slave agent JVMs, in which case you should use `Jenkins.getActiveInstance()` or 
* this is a class that is intended to be called from slave agent JVMs, in which case go away and rewrite this not to use the `Jenkins` class as we should not be loading the `Jenkins` class into slave agent JVMs...

Once you unpin that logic bomb, you will most likely be just using `Jenkins.getActiveInstance()` ;-)

I am expecting a follow-up PR

jglick: Obviously it is not to be called from slave agent JVMs. My only concern would be calls made during shutdown. I do not _think_ there should be any, since normally this method would be called in the context of an HTTP request handler thread (from an SCM webhook). We _do_ wait for all pending HTTP requests to complete before terminating, right @stephenc? …

PR #8, opened 2016-03-10
Clean up after PR#6

andresrc:
Follow up to https://github.com/jenkinsci/scm-api-plugin/pull/6

That PR upgrades the baseline but did not address `TODO` comments pending on such upgrade.

This PR fixes this. Some small additional cleanups (substitute deprecated findbugs annotations) are also included in the modified files.

Thanks @stephenc and @jglick for comments on the previous PR.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMHead.java
+++ b/src/main/java/jenkins/scm/api/SCMHead.java
@@ -38,7 +38,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import jenkins.model.Jenkins;
 import jenkins.model.TransientActionFactory;
 import jenkins.scm.api.actions.ChangeRequestAction;
 import org.kohsuke.stapler.export.Exported;
@@ -143,15 +142,12 @@ public class SCMHead implements Comparable<SCMHead>, Serializable {
     @Exported(name="actions")
     public List<? extends Action> getAllActions() {
         List<Action> actions = new ArrayList<Action>();
-        Jenkins j = Jenkins.getInstance(); // TODO 1.572+ ExtensionList.lookup
-        if (j != null) {
-            for (TransientActionFactory<?> taf : j.getExtensionList(TransientActionFactory.class)) {
-                if (taf.type().isInstance(this)) {
-                    try {
-                        actions.addAll(createFor(taf));
-                    } catch (Exception e) {
-                        LOGGER.log(Level.SEVERE, "Could not load actions from " + taf + " for " + this, e);
-                    }
+        for (TransientActionFactory<?> taf : ExtensionList.lookup(TransientActionFactory.class)) {
+            if (taf.type().isInstance(this)) {
+                try {
+                    actions.addAll(createFor(taf));
+                } catch (Exception e) {
+                    LOGGER.log(Level.SEVERE, "Could not load actions from " + taf + " for " + this, e);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


armfergom:
Looking at comments in previous PR (#6), this LGTM :bee:

src/main/java/jenkins/scm/api/SCMHead.java
```diff
@@ -38,12 +36,14 @@
...
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import jenkins.model.Jenkins;
 import jenkins.model.TransientActionFactory;
 import jenkins.scm.api.actions.ChangeRequestAction;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
```
jglick: I suspect @stephenc will bug this.

src/main/java/jenkins/scm/api/SCMHead.java
```diff
@@ -139,19 +139,16 @@ public String toString() {
...
-        Jenkins j = Jenkins.getInstance(); // TODO 1.572+ ExtensionList.lookup
-        if (j != null) {
-            for (TransientActionFactory<?> taf : j.getExtensionList(TransientActionFactory.class)) {
-                if (taf.type().isInstance(this)) {
-                    try {
-                        actions.addAll(createFor(taf));
-                    } catch (Exception e) {
-                        LOGGER.log(Level.SEVERE, "Could not load actions from " + taf + " for " + this, e);
-                    }
+        for (TransientActionFactory<?> taf : ExtensionList.lookup(TransientActionFactory.class)) {
```
jglick: Thanks, this was actually from my PR, filed before yours was merged IIRC.

src/main/java/jenkins/scm/api/SCMHead.java
stephenc: :bug:

stephenc:
JSR-305 is dormant, has been for a long while and shows no hope of ever producing an agreed set of annotations. Further more these annotations use javax. packages which it is not possible to use according to the Oracle Java binary licence, so applications can not use and ship these dependencies along with a JRE without violating the Oracle agreement.

> F. JAVA TECHNOLOGY RESTRICTIONS. You may not create, modify, or change the behavior of, or authorize your licensees to create, modify, or change the behavior of, classes, interfaces, or subpackages that are in any way identified as "java", "javax", "sun", “oracle” or similar convention as specified by Oracle in any naming convention designation.

The JSR-305 group has not defined any official releases according to its JSR page so the only implementations is a seemingly random implementation provided by the FindBugs team. Even if the Findbugs team were experts on the JSR (which some where) they are not official as there has been no vote and are not available from the JSR hompage - so the javax package name restriction still applies.

----

By using `javax` annotations bundled in jenkins.war we basically break the Oracle license in at least the docker image for Jenkins.

:bug:

BTW I am going to be filing a PR against Jenkins core that switches every annotation to a safe variant, e.g. the `edu.umd` one

andresrc:
@stephenc is there a safe non-deprecated variant?

stephenc:
Ignore the `@Deprecated`, there is [work on reversing that *or* getting the JSR revived](https://sourceforge.net/p/findbugs/bugs/1398/) /cc @jtnord

andresrc:
@stephenc Addressed and rebased.

andresrc:
@reviewbybees again esp. @stephenc

src/main/java/jenkins/scm/api/SCMHead.java
amuniz: Why is this change here?

andresrc: I was cleaning up `TODO`s and we are now 1.572+

amuniz: Ok, missed the comment.

jglick:
:bee:

recena:
@andresrc Maybe this should be included in `scm-api-1.2`, my mistake /cc @stephenc

PR #9, opened 2016-04-05
[JENKINS-33808] New API for support Item Categorization

recena:
[JENKINS-33808](https://issues.jenkins-ci.org/browse/JENKINS-33808)

@reviewbybees


src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -40,6 +40,12 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
         super(clazz);
     }
 
+    public abstract String getDescription();
```
jglick: :bug: missing Javadoc (3×).

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -40,6 +40,12 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+
+    public abstract String getCategoryId();
```
jglick: :bug: (3×) should probably be marked `@Nullable` in that a non-null return value is requested only if also returning non-null from `newInstance` (whose Javadoc should also be updated to note that creating a `newInstanceDetail` view is unnecessary if you implement `getDescription`).

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -40,6 +40,12 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+
+    public abstract String getIconFilePathPattern();
```
jglick: :bug: (3×) for introducing an `abstract` method. Should make it concrete and return null by default.

jglick:
:bug: define a default `newInstanceDetail.jelly`

recena:
@jglick some comments were addressed [here](https://github.com/jenkinsci/scm-api-plugin/pull/9/commits/4996fc14cbc6d1b36b91f4716954f1f79c76e932)

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
recena: @jglick What is the problem here?

jglick: You may not add `abstract` methods to an interface like this. It would not be compatible for existing subtypes.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,40 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
     /**
+     * A description of this kind of {@link SCMNavigator}. This description can contain HTML code but it is recommend
+     * to use text plain in order to avoid how it should be represented.
+     *
+     * @return A string, by default the value from newInstanceDetail view is taken.
+     *
+     * @since TODO
+     */
+    @Nonnull
+    public abstract String getDescription();
```
jglick: :bug: incompatible

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,40 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+     *
+     * @since TODO
+     */
+    @CheckForNull
+    public abstract String getIconFilePathPattern();
+
+    /**
      * Creates a default navigator, if there should be one displayed.
      * May be interpreted by {@code jenkins.branch.SpecificDescriptor}.
      * If returning non-null, you may also provide a {@code newInstanceDetail} view as per {@link TopLevelItemDescriptor}.
```
jglick: This should rather recommend returning the description from `getDescription`.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
recena: @jglick After this [commit](https://github.com/jenkinsci/scm-api-plugin/pull/9/commits/5b289d622c0619298c33d0d5c8bba0ec338a7b08) I have to work in this method.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,46 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
     }
 
     /**
+     * A description of this kind of {@link SCMNavigator}. This description can contain HTML code but it is recommend
```
jglick: recommended

src/main/resources/jenkins/scm/impl/SingleSCMNavigator/newInstanceDetail.jelly
```diff
@@ -0,0 +1,28 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core">
+  <j:out value="${it.description}"/>
```
jglick: :ant: why the override? This should be inherited from the parent.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,46 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+     * to use text plain in order to avoid how it should be represented.
+     *
+     * @return A string, by default the value from newInstanceDetail view is taken.
```
jglick: :ant: description of the default value is incorrect.

src/main/resources/jenkins/scm/api/Messages.properties
```diff
@@ -21,3 +21,5 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 #
+
+SCMNavigator.Description=Default implementation
```
jglick: :ant: meaningless to a user. Why not say something vague like

> A collection of SCM repositories.

that would stand some chance of being correct?

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,46 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+    /**
+     * Used to categorize this kind of {@link SCMNavigator}.
+     *
+     * @return A string with the category identifier.
+     *
+     * @since 1.2
+     */
+    @Nonnull
+    public String getCategoryId() {
+        return "uncategorized";
```
jglick: Should this not be `nested-projects`?

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,46 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+     */
+    @Nonnull
+    public String getDescription() {
+        return Messages.SCMNavigator_Description();
+    }
+
+    /**
+     * Used to categorize this kind of {@link SCMNavigator}.
+     *
+     * @return A string with the category identifier.
```
jglick: :ant: link to `TopLevelItemDescriptor.getCategoryId`

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,6 +42,46 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
...
+        return "uncategorized";
+    }
+
+    /**
+     * Represents a file path pattern to get the Item icon in different sizes.
+     *
+     * For example: plugin/plugin-shortname/images/:size/item.png, where {@code :size} represents the different
+     * icon sizes used commonly in Jenkins project: 16x16, 24x24, 32x32 or 48x48
+     *
+     * @return A string or null if it is not defined.
```
jglick: :ant: Javadoc link

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
jglick: :ant: Javadoc link

src/main/java/jenkins/scm/impl/SingleSCMNavigator.java
```diff
@@ -72,7 +72,12 @@ public void visitSources(SCMSourceObserver observer) throws IOException, Interru
 
         @Override
         public String getDisplayName() {
-            return "Single Repository";
+            return Messages.SingleSCMSource_DisplayName();
+        }
+
+        @Override
+        public String getDescription() {
+            return Messages.SingleSCMSource_Description();
```
jglick: :ant: pointless to override this, since it returns null from `newInstance` so this method will never be called.

jglick: (Which the Javadoc in the superclass should mention.)

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
recena: @jglick I don't like to write references to methods that only makes sense if you upgrade the baseline.

recena: @jglick In fact, the `mvn deploy` will fail generating the javadoc.

recena:
@jglick Do you agree with the changes?

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
jglick: Obviously you cannot use `@link`, but you _can_ use `@code` and leave a TODO comment to hyperlink it later.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -41,9 +42,51 @@ protected SCMNavigatorDescriptor(Class<? extends SCMNavigator> clazz) {
     }
 
     /**
+     * A description of this kind of {@link SCMNavigator}. This description can contain HTML code but it is recommended
+     * to use text plain in order to avoid how it should be represented.
```
jglick: :ant: Javadoc link

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -84,7 +84,9 @@ public String getIconFilePathPattern() {
     /**
      * Creates a default navigator, if there should be one displayed.
      * May be interpreted by {@code jenkins.branch.SpecificDescriptor}.
```
jglick: BTW this class name is obsolete.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -84,7 +84,9 @@ public String getIconFilePathPattern() {
...
-     * If returning non-null, you may also provide a {@code newInstanceDetail} view as per {@link TopLevelItemDescriptor}.
+     * If returning non-null, you may also provide a {@code newInstanceDetail} view as per {@link TopLevelItemDescriptor}
+     * and implement {@link SCMNavigatorDescriptor#getDescription()}.
```
jglick: Why would you implement `getDescription` if providing a `newInstanceDetail`? That description would not be used, AFAICT.

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
jglick: (resolved)

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
jglick: I think what you meant was that if you are returning non-null from `newInstance`, you should _either_ override `getDescription`, _or_ provide a `newInstanceDetail` with a more specialized appearance.

jglick: (:ant:)

jglick:
:bee:

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
jglick: …though IIRC in Jenkins 2, `newInstanceDetail` is only loaded for compatibility in case `getDescription` is not overridden, so we could perhaps ask the implementor to override `getDescription` and not even mention `newInstanceDetail`.

jglick: And it might be worth pointing out in the three new methods that they will not be used unless you return something from `newInstance`. In other words, if leaving `newInstance` to return null, you need not bother overriding the new methods.

jglick:
:bee:

PR #11, opened 2016-08-16
[JENKINS-32768] Round-tripping SingleSCMSource did not work

jglick:
[JENKINS-32768](https://issues.jenkins-ci.org/browse/JENKINS-32768)

@reviewbybees

PR #12, opened 2016-08-16
API for retrieving a fixed revision

jglick:
To be used from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/10. https://github.com/jenkinsci/git-plugin/pull/433 is the reference implementation.

Incorporates #11.

Note that current implementations of `SCMSource` only allow you to access the actual repository contents via the legacy `SCM` API, so use of the `build` method is key. In the future `SCMFileSystem` could be implemented (JENKINS-33273) which would allow you to directly observe files in a given revision without going through `SCM` at all, which would offer some advantages such as the ability to bypass `Run`s.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -34,6 +34,7 @@ import net.jcip.annotations.GuardedBy;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -305,6 +306,42 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
             }
         }, listener);
         return result.get();
+    }
+
+    /**
+     * Looks up suggested revisions that could be passed to {@link #fetch(String, TaskListener)}.
+     * There is no guarantee that all returned revisions are in fact valid, nor that all valid revisions are returned.
+     * Delegates to {@link #retrieveRevisions}.
+     * @param listener the task listener
+     * @return a possibly empty set of revision names suggested by the implementation
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @NonNull
+    public final Set<String> fetchRevisions(@CheckForNull TaskListener listener)
+            throws IOException, InterruptedException {
+        return retrieveRevisions(defaultListener(listener));
+    }
+
+    /**
+     * Looks up suggested revisions that could be passed to {@link #fetch(String, TaskListener)}.
+     * There is no guarantee that all returned revisions are in fact valid, nor that all valid revisions are returned.
+     * By default, calls {@link #retrieve(TaskListener)}, thus typically returning only branch names.
+     * @param listener the task listener
+     * @return a possibly empty set of revision names suggested by the implementation
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @NonNull
+    protected Set<String> retrieveRevisions(@NonNull TaskListener listener)
+            throws IOException, InterruptedException {
+        Set<String> revisions = new HashSet<String>();
+        for (SCMHead head : retrieve(listener)) {
+            revisions.add(head.getName());
+        }
+        return revisions;
     }
 
     /**
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -37,6 +37,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -258,6 +259,52 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
     protected SCMRevision retrieve(@NonNull SCMHead head, @NonNull TaskListener listener)
             throws IOException, InterruptedException {
         return fetch(SCMHeadObserver.select(head), listener).result();
+    }
+
+    /**
+     * Looks up a specific revision based on some SCM-specific set of permissible syntaxes.
+     * Delegates to {@link #retrieve(String, TaskListener)}.
+     * @param revision might be a branch name, a tag name, a cryptographic hash, a revision number, etc.
+     * @param listener the task listener (optional)
+     * @return a valid revision object corresponding to the argument, with a usable corresponding head, or null if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @CheckForNull
+    public final SCMRevision fetch(@NonNull String revision, @CheckForNull TaskListener listener)
+            throws IOException, InterruptedException {
+        return retrieve(revision, defaultListener(listener));
+    }
+
+    /**
+     * Looks up a specific revision based on some SCM-specific set of permissible syntaxes.
+     * The default implementation uses {@link #retrieve(SCMHeadObserver, TaskListener)}
+     * and looks for {@link SCMHead#getName} matching the argument (so typically only supporting branch names).
+     * @param revision might be a branch name, a tag name, a cryptographic hash, a revision number, etc.
+     * @param listener the task listener
+     * @return a valid revision object corresponding to the argument, with a usable corresponding head, or null if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @CheckForNull
+    protected SCMRevision retrieve(@NonNull final String revision, @NonNull TaskListener listener)
+            throws IOException, InterruptedException {
+        final AtomicReference<SCMRevision> result = new AtomicReference<SCMRevision>();
+        retrieve(new SCMHeadObserver() {
+            @Override
+            public void observe(SCMHead head, SCMRevision rev) {
+                if (head.getName().equals(revision)) {
+                    result.set(rev);
+                }
+            }
+            @Override
+            public boolean isObserving() {
+                return result.get() == null;
+            }
+        }, listener);
+        return result.get();
     }
 
     /**
--- a/src/main/java/jenkins/scm/impl/SingleSCMSource.java
+++ b/src/main/java/jenkins/scm/impl/SingleSCMSource.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
+import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItemDescriptor;
 import hudson.scm.NullSCM;
@@ -85,6 +86,13 @@ public class SingleSCMSource extends SCMSource {
         this.scm = scm;
     }
 
+    private synchronized void makeHead() {
+        if (head == null) {
+            head = new SCMHead(name);
+            revisionHash = new SCMRevisionImpl(head);
+        }
+    }
+
     /**
      * {@inheritDoc}
      */
@@ -93,11 +101,22 @@ public class SingleSCMSource extends SCMSource {
     protected synchronized void retrieve(@NonNull SCMHeadObserver observer,
                                          @NonNull TaskListener listener)
             throws IOException {
-        if (head == null) {
-            head = new SCMHead(name);
-            revisionHash = new SCMRevisionImpl(head);
-        }
+        makeHead();
         observer.observe(head, revisionHash);
+    }
+
+    /**
+     * Unconditionally returns the single nondeterministic revision.
+     * The {@code revision} argument is ignored.
+     * For the result of {@link #build(SCMHead, SCMRevision)} to be useful,
+     * ensure that {@link #scm} includes some parameterized fields,
+     * and that {@link Run#getEnvironment(TaskListener)} defines corresponding variables.
+     */
+    @NonNull
+    @Override
+    protected SCMRevision retrieve(String revision, TaskListener listener) throws IOException, InterruptedException {
+        makeHead();
+        return revisionHash;
     }
 
     /**
--- a/src/main/java/jenkins/scm/impl/SingleSCMSource.java
+++ b/src/main/java/jenkins/scm/impl/SingleSCMSource.java
@@ -28,7 +28,6 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
-import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItemDescriptor;
 import hudson.scm.NullSCM;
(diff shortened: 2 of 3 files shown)


src/main/java/jenkins/scm/impl/SingleSCMSource.java
```diff
@@ -93,14 +109,25 @@ public SingleSCMSource(String id, String name, SCM scm) {
...
+        makeHead();
         observer.observe(head, revisionHash);
     }
 
     /**
+     * Unconditionally returns the single nondeterministic revision.
+     * The {@code revision} argument is ignored.
+     * For the result of {@link #build(SCMHead, SCMRevision)} to be useful,
+     * ensure that {@link #scm} includes some parameterized fields,
+     * and that {@link Run#getEnvironment(TaskListener)} defines corresponding variables.
```
jglick: See `LibraryAdderTest.usingInterpolation`, relying on `LibrariesAction.LibraryEnvironment`.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -261,6 +262,52 @@ protected SCMRevision retrieve(@NonNull SCMHead head, @NonNull TaskListener list
...
+     * @param revision might be a branch name, a tag name, a cryptographic hash, a revision number, etc.
+     * @param listener the task listener
+     * @return a valid revision object corresponding to the argument, with a usable corresponding head, or null if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @CheckForNull
+    protected SCMRevision retrieve(@NonNull final String revision, @NonNull TaskListener listener)
+            throws IOException, InterruptedException {
```
jglick: A possible future extension would be a way to list all suggested revisions (not things like commit hashes but at least tags and so on). See comment in `LibraryConfiguration.DescriptorImpl`. Could be added compatibly.

pom.xml
```diff
@@ -29,7 +29,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.3</version>
+    <version>2.12</version>
```
svanoort: Any particular reason for the parent pom bump?

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -261,6 +262,52 @@ protected SCMRevision retrieve(@NonNull SCMHead head, @NonNull TaskListener list
...
+     * @param revision might be a branch name, a tag name, a cryptographic hash, a revision number, etc.
+     * @param listener the task listener (optional)
+     * @return a valid revision object corresponding to the argument, with a usable corresponding head, or null if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @CheckForNull
+    public final SCMRevision fetch(@NonNull String revision, @CheckForNull TaskListener listener)
+            throws IOException, InterruptedException {
```
svanoort: What if the revision format is invalid?  We will end up with an IOException, but would it make sense to throw a more specific exception type than that (maybe InvalidSCMRevision)? 

This would also let us do some prevalidation before issuing network calls to we don't try with an invalid (or potentially security-compromising) revision specifier.

src/main/java/jenkins/scm/api/SCMSource.java
svanoort: Per above: option to throw a custom exception if revision format is invalid, perhaps?  Otherwise we just get a null.  

This has much greater benefits for custom implementations (GitSCM, I'm looking at you!) even if it is not as helpful for the default impl.

pom.xml
svanoort: Note for other reviewers: from #11 (explained there, and incorporated into this).

svanoort:
🐝  though I think an option to throw a more explicit custom exception for invalid revisions/revision formats would be very helpful.  

Two immediate use cases I can see:
* Requesting a revision that trivial validation indicates is invalid (empty, illegal characters, too long).  This lets you display a more helpful user message and avoids needless network communications.
* Trying to retrieve a revision that the SCM rejects as invalid (for reasons other than "does not exist").  This makes it easier to do error handling when interacting with SCM.

src/main/java/jenkins/scm/api/SCMSource.java
jglick: > What if the revision format is invalid? We will end up with an IOException

It will return null.

src/main/java/jenkins/scm/api/SCMSource.java
jglick: > we just get a null

Yes but `listener` will receive diagnostics.

In the case of `GitSCM`, the plugin really has no idea ahead of time what revision  formats would be valid. It calls `git rev-parse` and either that produces something or it does not.

jglick: See `AbstractGitSCMSourceTest.retrieveRevision`’s `stderr` for example:

```
…
 > git rev-parse nonexistent^{commit} # timeout=10
 > git rev-parse origin/nonexistent^{commit} # timeout=10
Command "git rev-parse nonexistent^{commit}" returned status code 128:
stdout: nonexistent^{commit}

stderr: fatal: ambiguous argument 'nonexistent^{commit}': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

Command "git rev-parse origin/nonexistent^{commit}" returned status code 128:
stdout: origin/nonexistent^{commit}

stderr: fatal: ambiguous argument 'origin/nonexistent^{commit}': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
…
```

jglick:
> Trying to retrieve a revision that the SCM rejects as invalid (for reasons other than "does not exist")

Perhaps some SCM implementations could benefit. I do not think any implementation of that is possible in the `git` plugin as written, nor in the `hg` plugin. The `subversion` plugin may be able to get that information somehow via SVNKit.

Anyway I am not sure what the caller would do with this information, exactly. Either it gets a valid revision, or it does not, and the log shows details that a user can read.

> avoids needless network communications

An implementation is already free to return null without doing I/O if it can detect that a revision argument is syntactically malformed.

src/main/java/jenkins/scm/api/SCMSource.java
jglick: > potentially security-compromising

An implementation is of course on the hook to ensure that security is not compromised by a caller passing a malicious revision argument. In the case of in-JVM connectors this is likely a nonissue. For those using external processes it is generally handled by use of injection-proof calls like `Runtime.exec(String[])`.

src/main/java/jenkins/scm/api/SCMSource.java
svanoort: So, in this case it's quite feasible to programmatically implement some of these rules, before handing off to rev-parse (they're complex, but usable). 

Listener isn't an option AFAIK because if you have to parse the stderr output directly to offer programmatic handling, that opens up a whole nasty can of worms: better to do some prevalidation so you can show UI validation rules rather than handing off straight to rev-parse.

src/main/java/jenkins/scm/api/SCMSource.java
svanoort: How is Runtime.exec(String[]) injection-proof?  On its own there's nothing stopping you from doing '|| sudo cat /etc/shadow' as one of the command strings, is there?

svanoort:
The more I think about it, the more I'm convinced that this is incomplete without some sort of trivial handling of invalid revision that is separate from a null return (does not exist / rejected) or IOException (can't communicate, or some other lower-level issue). 

>  I do not think any implementation of that is possible in the git plugin as written, nor in the hg plugin.

There are rules, they're just complex, but the rev-parse docs includes them: https://www.kernel.org/pub/software/scm/git/docs/git-rev-parse.html

Many of the more complex rules are in fact safe to reject because they require being on a currently checked out branch (ex: the relative revs).

svanoort:
I can't mark it as a bug but I consider this one a very hard :ant: if we're adding the ability to fetch revisions by arbitrary string without providing some way to validate (preferably with an SCM-overrideable validation function used, with signature like):

```java
/** Override for SCM-specific validation of revision names */
public boolean isRevisionFormatValid(@NonNull String revision) {
   return !revision.isEmpty(); 
}
```

Stipulation that these validation APIs shouldn't trigger actual I/O.

stephenc:
:+1: from a quick glimpse (and you owe me for interruption ;-) )

jglick:
> without providing some way to validate

The problem I have with `isRevisionFormatValid` is that there is no really feasible implementation for the reference implementation (`AbstractGitSCMSource`), without duplicating pages of complex parser rules; and there is no candidate caller for the API either, since `LibraryAdder` gets along fine without it. By the YAGNI principle we should not be defining an API without a demonstrated use case, when there is nothing that I can see blocking such an API from being added retroactively if one arises.

The suggestion to throw `InvalidSCMRevision` from `retrieve` is a little trickier since this would mean changing the semantics of an existing API, though if it `extends IOException` it would be harmless since callers could just opt in to handling the more specific exception differently.

svanoort:
>  is that there is no really feasible implementation for the reference implementation (AbstractGitSCMSource)

You could easily implement a subset of this and get reasonable results (at least rejecting the majority of the invalid results):  http://stackoverflow.com/a/3651867/95122

> The suggestion to throw InvalidSCMRevision from retrieve is a little trickier since this would mean changing the semantics of an existing API, though if it extends IOException it would be harmless since callers could just opt in to handling the more specific exception differently.

This would be a reasonable back-compatible compromise to me.  Not ideal, but it would serve the purpose well enough and provide the extra information for handlers to consume.

jglick:
> implement a subset of this

Well the referenced answer refers to branch names. But we support tag names, which certainly may contain `.`.

jglick:
BTW in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/10 I have already implemented form validation using the existing signature.

(I also added a method to _enumerate_ revisions, which unfortunately I cannot use for auto-completion due to limitations of the JavaScript plumbing in Jenkins forms.)

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -308,6 +309,42 @@ public boolean isObserving() {
...
+     * There is no guarantee that all returned revisions are in fact valid, nor that all valid revisions are returned.
+     * By default, calls {@link #retrieve(TaskListener)}, thus typically returning only branch names.
+     * @param listener the task listener
+     * @return a possibly empty set of revision names suggested by the implementation
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since FIXME
+     */
+    @NonNull
+    protected Set<String> retrieveRevisions(@NonNull TaskListener listener)
```
svanoort: Seems reasonable.

svanoort:
I like the addition for enumerating revisions (more precisely, *head* revisions, as expected for SCMHead/SCMRevision).  

> Well the referenced answer refers to branch names. But we support tag names, which certainly may contain ..

Implementation details, where we're defining an API here (impl discussion is for the other PR).  Why does our validation rule have to be exhaustive?  

* To my mind, it's sufficient to filter out the majority of obviously-invalid results, and the SCM itself can still kick back more complex violations.  
* A custom Exception (or in perhaps IOException) class provides a tidy way to bubble up deeper validation failures that can't be trivially rejected by the previous rule.

Note: none of this would matter if we weren't accepting arbitrary string input, and restricting valid strings to only heads.  That might also be a reasonable default validation rule (using your new enumeration API, perhaps).

jglick:
> addition for enumerating revisions (more precisely, head revisions

These are not, in general, “head revisions”. For the Git implementation, it includes tags. The _default_ implementation returns heads.

> obviously-invalid results

Not sure what you mean by that.

> restricting valid strings to only heads

That is precisely what we do _not_ want to do, because that would prevent us from resolving tags, commit hashes, or stranger beasts. Using the existing released API, we could already resolve branches.

svanoort:
> Not sure what you mean by that.

Things that can obviously be discarded as invalid. 

> That is precisely what we do not want to do, because that would prevent us from resolving tags, commit hashes, or stranger beasts. Using the existing released API, we could already resolve branches.

Implementations are going to have to define their own rules anyway.  The other alternative would be to only filter out empty strings as invalid revisions and let implementations extend it. 

My basic reasoning here is that we need more robust common errors/validation in general if we're adding the capability to check out arbitrary revisions.  Otherwise each SCM impl is going to roll their own mechanism & methods, and there's not going to be an easy way to handle these failure cases consistently across SCMs using the common API.

jglick:
> Whitespace. Unicode characters where you're expecting hex. Reserved characters

As far as I know most of these things are potentially valid (if unwise) tag names.

> filter out empty strings as invalid revisions

Tests in the downstream https://github.com/jenkinsci/git-plugin/pull/433 shows that even this is not really necessary, though that plugin could print a specialized error message for that case if it cared.

> each SCM impl is going to roll their own mechanism & methods

As they should: each SCM has its own meanings for revision strings.

> there's not going to be an easy way to handle these failure cases consistently across SCMs

How so? If null is returned, the `TaskListener` gets errors. Which we successfully show in the UI in the downstream https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/10.

svanoort:
> As they should: each SCM has its own meanings for revision strings.

They shouldn't have completely separate error handling APIs though -- that defeats the point of having a common API. 

> If null is returned, the TaskListener gets errors.

Already addressed above -- better than swallowing the error, but really only good for log-and-fail handling and no more.  Limits the usefulness of this API (which I want to be clear: is really quite useful in many, many ways).   

We may have to agree to disagree here -- the current implementation will *technically* work, it's just that in my opinion it sets us up for significant pain down the road.  If I didn't flag that design problem, I would not be doing my job as a reviewer here.

jglick:
> better than swallowing the error, but really only good for log-and-fail handling and no more

Well, log-and-fail handling is the only use case that has been identified so far. Either the user supplied a string which the SCM plugin can resolve to a particular exact revision of the repository, so we proceed; or they did not, and we display whatever messages the SCM offers. This works both at runtime in the build log, and at configuration time in form validation.

> sets us up for significant pain down the road

What would that pain be? If some use case develops for, say, displaying distinct UI controls (not just distinct text) based on different generic categories of problem with revisions, then we would define some `public` subtypes of `IOException` in `scm-api` modeling those categories, `catch` them from callers, and update important SCM implementations to `throw` them when appropriate. Nothing would be broken for old callers mixed with new SCMs, or new callers mixed with old SCMs, or old callers mixed with old SCMs.

abayer:
:bee:

src/main/java/jenkins/scm/api/SCMSource.java
jglick: The overload taking a `String[]` bypasses the shell’s parser, so it will wind up doing the same thing as

```
$ git rev-parse '|| sudo cat /etc/shadow^commit'
|| sudo cat /etc/shadow^commit
fatal: ambiguous argument '|| sudo cat /etc/shadow^commit': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

svanoort:
> define some public subtypes of IOException in scm-api

@jglick If you provide a subtype of IOException such as "InvalidRevisionException" or "RevisionParsingException" and add Javadocs to the new API to indicate its use, that would be enough to consider the condition satisfied, I think (have mentioned that before).

In any case you have my 🐝  already, so this is not a blocker for merge/release.

jglick:
I remain reluctant to define, and thus commit to in perpetuity, an exception type which

* no written implementation would throw
* no written caller would catch
* I am reasonably confident could be added compatibly if and when the first two conditions change

rsandell:
:bee:

PR #13, opened 2016-08-25
Forgot to call getSCMDescriptors

jglick:
Corrects a regresson in #11 whereby, for example, **None** was displayed as an option.

@reviewbybees


abayer:
:bee:

PR #14, opened 2016-09-17
Preliminary high level documentation in the README

MarkEWaite:
Describes the data model and data flow, with hyperlinks to reference
implementations.

Special thanks to @p4paul for his help, and to Jesse Glick for his
description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,3 +27,63 @@ To test in a local Jenkins instance
     mvn hpi:run
 
   [wiki]: http://wiki.jenkins-ci.org/display/JENKINS/SCM+API+Plugin
+
+# SCM API Data Model
+
+## SCMSource
+
+[SCMSource](src/main/java/jenkins/scm/api/SCMSource.java)
+[(doc)](http://javadoc.jenkins.io/plugin/scm-api/jenkins/scm/api/SCMSource.html)
+implementations provide a configuration form (as a
+[Descriptor](http://javadoc.jenkins-ci.org/hudson/model/Descriptor.html))
+for users, and define the methods for multi-branch and other advanced
+SCM concepts in Jenkins.
+
+The most crucial methods an
+[SCMSource](src/main/java/jenkins/scm/api/SCMSource.java) must
+implement are:
+
+* [retrieve(HeadObserver, TaskListener)](https://github.com/jenkinsci/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L153) - provides HeadObserver with the Set(name, hash) of revisions to consider
+* [retrieve(SCMHead, TaskListener)](https://github.com/MarkEWaite/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L222) - returns a single revision
+* [retrieve(String, TaskListener)](https://github.com/MarkEWaite/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L282) - returns a single revision identified by the string value
+
+## SCMHead
+
+[SCMHead](src/main/java/jenkins/scm/api/SCMHead.java)
+[(doc)](http://javadoc.jenkins.io/plugin/scm-api/jenkins/scm/api/SCMHead.html)
+is a named tree (subversion branch, mercurial branch, git branch,
+perforce branch/stream, etc.), identified by name
+
+## SCMRevision
+
+[SCMRevision](src/main/java/jenkins/scm/api/SCMRevision.java)
+[(doc)](http://javadoc.jenkins.io/plugin/scm-api/jenkins/scm/api/SCMRevision.html)
+is a specific change in the tree (subversion revision, mercurial hash,
+git sha, perforce change, etc.) identified by a version control
+specific object.
+
+## SCMStatus
+
+SCMStatus implementations (like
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,4 +26,64 @@ To test in a local Jenkins instance
...
+[(doc)](http://javadoc.jenkins.io/scm-api/jenkins/scm/api/SCMSource.html#build-jenkins.scm.api.SCMHead-jenkins.scm.api.SCMRevision)
+method returns a reference to your SCM implementation as constructed
+from user provided values (e.g. jelly provided values).  Refer to
+[mercurial](https://github.com/jenkinsci/mercurial-plugin/blob/4e9fcd772cd4a4c1264e19a175476b5e48cd1c94/src/main/java/hudson/plugins/mercurial/MercurialSCMSource.java#L142) as a reference implementation.
+
+The [getCriteria()](https://github.com/jenkinsci/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L125)
+[(doc)](http://javadoc.jenkins.io/scm-api/jenkins/scm/api/SCMSource.html#getCriteria)
+method provides criteria (such as the presence of a Jenkinsfile in the
+root of the repository branch) used to select which branches should
+have jobs created for them.
```
p4paul: SCMSourceCriteria Implementations in:
[AbstractGitSCMSource](https://github.com/jenkinsci/git-plugin/blob/1b555c8d995413160f3eea80b3f6926e7b411369/src/main/java/jenkins/plugins/git/AbstractGitSCMSource.java#L296)

jglick:
Probably needs to overhauled in light of #17.

stephenc:
I am inclined to close this in favour of the documentation in #17

stephenc:
I think some of the content here is plain wrong. The others is perhaps duplicated with the new consumer and implementation docs.

If you still think an overview could help then we can see about reworking this to fix the wrong content... otherwise we can just close this.

@MarkEWaite WDYT?

README.md
```diff
@@ -26,4 +26,66 @@ To test in a local Jenkins instance
...
+implementations provide a configuration form (as a
+[Descriptor](http://javadoc.jenkins-ci.org/hudson/model/Descriptor.html))
+for users, and define the methods for multi-branch and other advanced
+SCM concepts in Jenkins.
+
+The most crucial methods an
+[SCMSource](src/main/java/jenkins/scm/api/SCMSource.java) must
+implement are:
+
+* [retrieve(HeadObserver, TaskListener)](https://github.com/jenkinsci/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L153) - provides HeadObserver with the Set(name, hash) of revisions to consider
```
stephenc: Nope. Recommendation is now to implement `retrieve(criteria,observer,event,listener)` and this method is now abstract, so should be the one that IDEs suggest implementing

README.md
```diff
@@ -26,4 +26,66 @@ To test in a local Jenkins instance
...
+* [retrieve(SCMHead, TaskListener)](https://github.com/MarkEWaite/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L222) - returns a single revision
```
stephenc: if you have implemented `getIncludes()` support as recommended in the implementation guide, then overriding this method is likely unnecessary

README.md
```diff
@@ -26,4 +26,66 @@ To test in a local Jenkins instance
...
+* [retrieve(String, TaskListener)](https://github.com/MarkEWaite/scm-api-plugin/blob/c06684f6e252bd4f8ef99ab8ec6eefceed8dce45/src/main/java/jenkins/scm/api/SCMSource.java#L282) - returns a single revision identified by the string value
```
stephenc: Implementing this is not required, but if you want a more performant implementation it can be helpful depending on the cost of getting the list of name/revision pairs

README.md
```diff
@@ -26,4 +26,66 @@ To test in a local Jenkins instance
...
+
+## SCMHead
+
+[SCMHead](src/main/java/jenkins/scm/api/SCMHead.java)
```
stephenc: Also need to note that the `SCMHead` is treated like a key, so the equals contract is special. It should be treated as a "string" with type info rather than a store of data. The mixin interfaces can put additional constraints on equality

README.md
```diff
@@ -26,4 +26,66 @@ To test in a local Jenkins instance
...
+[(doc)](http://javadoc.jenkins.io/plugin/scm-api/jenkins/scm/api/SCMHead.html)
+is a named tree (subversion branch, mercurial branch, git branch,
+perforce branch/stream, etc.), identified by name
+
+## SCMRevision
+
+[SCMRevision](src/main/java/jenkins/scm/api/SCMRevision.java)
+[(doc)](http://javadoc.jenkins.io/plugin/scm-api/jenkins/scm/api/SCMRevision.html)
+is a specific change in the tree (subversion revision, mercurial hash,
```
stephenc: specific _version of_ the tree not _change in_

README.md
```diff
@@ -26,4 +26,66 @@ To test in a local Jenkins instance
...
+git sha, perforce change, etc.) identified by a version control
+specific object.
+
+## SCMStatus
+
+SCMStatus implementations (like
```
stephenc: I provide a new recommended path for this in the implementation guide. I recommend new SCM implementations do not follow this path, rather consolidate events through the SCM API's event system

MarkEWaite:
Let's close this pull request.  The content in #17 is preferred and I don't think the overview is useful enough to rework it to remove the mistakes.

PR #15, opened 2016-10-05
Extension point for exposing default branches

michaelneale:
Ticket: [JENKINS-38718](https://issues.jenkins-ci.org/browse/JENKINS-38718)

I believe this probably belongs in scm-api, as it is the common denominator that various systems share (github will have to implement this, but not git, as I don't think git has the concept of a default branch, but github does, for example). 

Just opening for discussion...

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMDefaultBranch.java
+++ b/src/main/java/jenkins/scm/api/SCMDefaultBranch.java
@@ -1,0 +1,18 @@
+package jenkins.scm.api;
+
+import hudson.ExtensionPoint;
+
+/**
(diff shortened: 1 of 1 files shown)


jglick:
Need at least one (preferably multiple) reference implementations, and at least one reference caller, as downstream PRs demonstrating end-to-end usage.

michaelneale:
agree... @jglick the main use for this is GUI related, are you aware of other needs for this?

stephenc:
:bug: I would prefer if the default branch was handled via the new Categorization API

michaelneale:
yes agree, closing this for that reason

PR #17, opened 2016-10-28
[FIXED JENKINS-39355] Various API improvements

stephenc:
See [JENKINS-39355](https://issues.jenkins-ci.org/browse/JENKINS-39355)

@reviewbybees 

Notes:
- There are no implementations of `SCMFileSystem` so we can safely break backwards compatibility.
- Examples of use PRs are in progress


jglick:
> Examples of use PRs are in progress

Still? Should be more meaningful to conduct a review of the API when there are some concrete implementations showing the new features in action.

jglick:
> we can safely break backwards compatibility

Would be wise to bump the version to 2.0 as an indication of this.

src/main/java/jenkins/scm/api/SCMSourceCategory.java
```diff
@@ -0,0 +1,188 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import jenkins.scm.impl.UncategorizedSCMSourceCategory;
+import org.jvnet.localizer.Localizable;
+
+/**
+ * Base class for categories of {@link SCMSource}, for example: repository, depot, vob, etc.
```
jglick: `VOB`

src/main/java/jenkins/scm/api/SCMSourceCriteria.java
```diff
@@ -51,7 +51,7 @@
     /**
      * A probe for a branch candidate. Inspectors can tell whether a file path exists.
      */
-    static abstract class Probe implements Serializable {
+    abstract class Probe implements Serializable {
```
jglick: Did you really need to make this non`static`? Will break serial form for sure, and maybe binary compatibility too.

If you really need a reference to `SCMSourceCriteria` for some reason, it is better to introduce an explicit, nullable field with an injector method, so that you can consciously control compatibility issues.

src/main/java/jenkins/scm/api/SCMSourceDescriptor.java
```diff
@@ -44,6 +48,15 @@
 public abstract class SCMSourceDescriptor extends Descriptor<SCMSource> implements IconSpec {
 
     /**
+     * The set of {@link SCMHeadCategory} singletons for this type of {@link SCMSource}
+     *
+     * @see #getCategories()
+     * @since FIXME
+     */
+    @GuardedBy("this")
+    protected transient Set<SCMHeadCategory> categories;
```
jglick: `private`

src/main/java/jenkins/scm/api/SCMSourceObserver.java
```diff
@@ -24,34 +24,59 @@
...
      * Provides a way of reporting progress.
+     *
      * @return a logger
      */
     @NonNull
     public abstract TaskListener getListener();
 
     /**
+     * Returns the subset of project names that this observer is interested in or {@code null} if
+     * interested in all project names.
```
jglick: What are “project names” in this context, exactly? `SCMHead.name`s? Encoded?

src/main/java/jenkins/scm/impl/TagSCMHeadCategory.java
```diff
@@ -0,0 +1,91 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+/*
+ * The MIT License
```
jglick: duplicated

src/main/java/jenkins/scm/impl/UncategorizedSCMHeadCategory.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Constructs a {@link UncategorizedSCMHeadCategory} using the default naming.
+     */
+    public UncategorizedSCMHeadCategory() {
+        super(Messages._UncategorizedSCMHeadCategory_DisplayName());
+    }
+
+    /**
+     * Constructs a {@link UncategorizedSCMHeadCategory} with customized naming. Use this constructor when the generic
+     * naming is not appropriate terminology for the specific {@link SCMSource}'s naming of branches
```
jglick: Does not make sense. Why would you have a custom naming for something which by its nature represents things you do not know about?

src/main/java/jenkins/scm/impl/UncategorizedSCMSourceCategory.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Constructs a {@link UncategorizedSCMSourceCategory} using the default naming.
+     */
+    public UncategorizedSCMSourceCategory() {
+        super(Messages._UncategorizedSCMSourceCategory_DisplayName());
+    }
+
+    /**
+     * Constructs a {@link UncategorizedSCMSourceCategory} with customized naming. Use this constructor when the generic
+     * naming is not appropriate terminology for the specific {@link SCMNavigator}'s naming of repositories.
```
jglick: Ditto.

src/test/java/jenkins/scm/api/MessagesTest.java
```diff
@@ -0,0 +1,138 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+/*
+ * The MIT License
```
jglick: several duplicates

src/test/java/jenkins/scm/api/MessagesTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
+import org.junit.runner.RunWith;
+
+import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.Assert.assertThat;
+
+/**
+ * The code coverage for the generated Messages class doesn't matter, so this test stops them from dragging the coverage
+ * down and results in a clearer measure of code coverage for the module.
```
jglick: Please delete, this is nonsense. Of course any plugin will have code which does not get coverage; that is something for humans and/or tools to deal with.

src/main/java/jenkins/scm/impl/ChangeRequestSCMHeadCategory.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Standard category for {@link ChangeRequestSCMHead} instances.
+ *
+ * @since FIXME
+ */
+public final class ChangeRequestSCMHeadCategory extends SCMHeadCategory {
+    /**
+     * Constructs a {@link ChangeRequestSCMHeadCategory} using the default naming.
+     */
+    public ChangeRequestSCMHeadCategory() {
```
jglick: Why is this in the `impl` package? Would not an SCM implementation wish to return this among others?

src/test/java/jenkins/scm/api/SCMHeadObserverTest.java
```diff
@@ -0,0 +1,138 @@
...
+        instance.observe(head1, revision1);
+        assertThat("Still observing before match", instance.isObserving(), is(true));
+        instance.observe(head2, revision2);
+        assertThat("Stops observing after selected observation", instance.isObserving(), is(false));
+        assertThat(instance.result(), is(revision2));
+    }
+
+    @Test
+    public void filter() throws Exception {
+
```
jglick: …?

src/test/java/jenkins/scm/impl/MessagesTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
+import org.junit.runner.RunWith;
+
+import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.Assert.assertThat;
+
+/**
+ * The code coverage for the generated Messages class doesn't matter, so this test stops them from dragging the coverage
+ * down and results in a clearer measure of code coverage for the module.
```
jglick: Ditto.

src/test/java/jenkins/scm/impl/SingleSCMNavigatorTest.java
```diff
@@ -0,0 +1,124 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.when;
+
+public class SingleSCMNavigatorTest {
+    @Test
+    public void getName() throws Exception {
+        Random entropy = new Random();
+        String name = "foo-" + entropy.nextLong();
```
jglick: Why does this need to be randomized?

src/test/java/jenkins/scm/impl/TagSCMHeadCategoryTest.java
```diff
@@ -0,0 +1,95 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+/*
+ * The MIT License
```
jglick: dupe

src/test/java/jenkins/scm/impl/UncategorizedSCMHeadCategoryTest.java
```diff
@@ -0,0 +1,120 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+/*
+ * The MIT License
```
jglick: dupe

src/test/java/jenkins/scm/impl/UncategorizedSCMSourceCategoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+/*
+ * The MIT License
```
jglick: Yow, fix your IDE!

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
+== Implementation Guide
```
jglick: Link to this from `README.md`.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+The SCM API consists of a number of extension points:
+
+* The original `hudson.scm.SCM` extension point is responsible for:
+
+** performing checkouts from source control into a `FilePath`
+
+** calculating the changelogs between revisions
+
+** constructing the changelog parser
```
jglick: Though probably in the future we will need `SCMSource` to take over at least the second function; something like

```java
public abstract ChangeLogSet<?> diff(SCMRevision start, SCMRevision end);
```

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+* Git has a fairly direct mapping of concepts
+
+** `jenkins.scm.api.SCMHead` would represent the branches and tags in an individual Git repository
+
+** `jenkins.scm.api.SCMRevision` would represent the Git commit hash of each individual commit. Because Git commits are immutable, each `SCMRevision` is deterministic as if you ask to check out the same hash you will always get the same contents.
+
+** `jenkins.scm.api.SCMSource` would represent a single Git repository.
+
+** `jenkins.scm.api.SCMNavigator` would represent a collection of Git repositories - most likely all on the same server
```
jglick: To be concrete, may as well mention a GitHub organization as an example.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
+== Implementation Guide
+
+This document provides the guidelines for a correct optimal implementation of the SCM API.
```
jglick: @KostyaSha IIRC you were asking for something like this.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+** `jenkins.scm.api.SCMSource` would represent a single Git repository.
+
+** `jenkins.scm.api.SCMNavigator` would represent a collection of Git repositories - most likely all on the same server
+
+* Subversion can have a fairly direct mapping of concepts analogous to the Git mapping above.
+Alternatively some oganizations will use a single Subversion repository to hold multiple projects under which each project would have the familiar `trunk/branches/tags` structure. If we consider this style of Subversion use then:
+
+** `jenkins.scm.api.SCMHead` would be mapped against a specific project and be one of: `trunk`, `branches/_name_` or `tags/_name_`.
+
+** `jenkins.scm.api.SCMRevision` would be mapped to the global Subversion revision number against which the specific `SCMHead` was last modified. Each `SCMRevision` is deterministic as if you ask to check out the trunk/branch.tag at the same revision you will always get the same contents.
```
jglick: Helpful to clarify that this corresponds to something like `http://server/project/trunk/@12345`.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+** `jenkins.scm.api.SCMSource` would represent a single project withi the Subversion repository.
```
jglick: typo

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+** `jenkins.scm.api.SCMNavigator` would represent the collection of projects in the root of the Subversion repository.
+
+* CVS, one mapping of the CVS concepts would be:
+
+** `jenkins.scm.api.SCMHead` would represent a branch in a module.
+
+** `jenkins.scm.api.SCMRevision` would represent a timestamp. Sadly the timestamp is not deterministic as an immediate checkout may miss some files that are still being committed to the repository.
```
jglick: Timestamp as well as the branch name, since a revision always refers to a specific head.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+** `jenkins.scm.api.SCMRevision` would be mapped to the global Subversion revision number against which the specific `SCMHead` was last modified. Each `SCMRevision` is deterministic as if you ask to check out the trunk/branch.tag at the same revision you will always get the same contents.
+
+** `jenkins.scm.api.SCMSource` would represent a single project withi the Subversion repository.
+
+** `jenkins.scm.api.SCMNavigator` would represent the collection of projects in the root of the Subversion repository.
+
+* CVS, one mapping of the CVS concepts would be:
+
+** `jenkins.scm.api.SCMHead` would represent a branch in a module.
```
jglick: Or `HEAD` or a tag.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+* Transitioned the criteria matching from `false` to `true`;
+* Maintained the criteria matching in its current state; or
+* Transitioned the criteria matching from `true` to `false`.
+
+Thus *every* SCM API consuming plugin that listens for a `jenkins.scm.api.SCMEvent` will need to validate that event *against the criteria*.
+This revalidation would be necessary even if we received a signed trusted event from the backing source control system.
+This is why:
+
+* The `jenkins.scm.api.SCMEvent` is always considered `Untrusted` - consuming plugins will be doing revalidation anyway, so it is simpler for consuming plugins to treat everything from an event as untrusted.
```
rsandell: I read this paragraph three times and I'm still fuzzy on the details of what you are trying to say.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+** `jenkins.scm.api.SCMNavigator` would represent the collection of projects in the root of the Subversion repository.
+
+* CVS, one mapping of the CVS concepts would be:
+
+** `jenkins.scm.api.SCMHead` would represent a branch in a module.
+
+** `jenkins.scm.api.SCMRevision` would represent a timestamp. Sadly the timestamp is not deterministic as an immediate checkout may miss some files that are still being committed to the repository.
++
+NOTE: An alternative deterministic `jenkins.scm.api.SCMRevision` would be a list of all the files and their individual revisions, but this would be prohibitive in terms of storage cost and is non-trivial for a user to replicate from their own workspace and consequently the timestamp would be preferred.
+Ignoring the problems of time sychronization in a distributed system, we could assume that any timestamp more than _X_ minutes in the past should be deterministic in order to allow the consuming plugins to make optimizations that are only possible with deterministic revisions.
```
jglick: A CVS timestamp is actually deterministic down to the granularity of one second. It is true that a developer might have been in the process of running a commit which took several minutes (this was in fact common when I worked on the NetBeans megarepo prior to Mercurial conversion), so the snapshot might be _incomplete_ (for example, uncompilable); but it is still very nearly _deterministic_ in that two different checkouts of the same `SCMRevision` should produce the same files—possibly differing only in some changes delivered within the span of one second, and even then only if one of the checkouts occurred _simultaneously_ with the commit, during that same second interval (checkouts happening later should agree on whether a given timestamp was before or after).

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+** `jenkins.scm.api.SCMSource` would represent an individual module on the CVS server.
+
+** `jenkins.scm.api.SCMNavigator` would represent the collection of modules available from a single CVS server.
+
+Implementers are free to map the concepts to their own SCM system as they see fit, but the recommendation is to try to keep close to the principles of mapping outlined in the above examples.
```
jglick: @p4paul please contribute a description of how Perforce would map, since it has a rather more complicated set of options.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+The concepts we have covered so far determine how Jenkins plugins can drive interactions with the SCM system.
+While a Jenkins driven interaction with an SCM is sufficient for enabling advanced SCM functionality such as that provided in the https://wiki.jenkins-ci.org/display/JENKINS/Branch+API+Plugin[Branch API plugin], it does not lend to a good user experience as Jenkins would be required to continually poll the backing SCM to establish if there are any changes.
```
jglick: Whoa, slow down. Before people can implement events, they need to have a working plugin doing the basics, and this is not trivial. We need an explanation of how you can implement simple whole-repo branch scanning and whole-org repo scanning.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+** enumerating potential `jenkins.scm.api.SCMSource` instances
+
+Some examples of how these concepts may be mapped into well known source control systems:
+
+* Git has a fairly direct mapping of concepts
+
+** `jenkins.scm.api.SCMHead` would represent the branches and tags in an individual Git repository
+
+** `jenkins.scm.api.SCMRevision` would represent the Git commit hash of each individual commit. Because Git commits are immutable, each `SCMRevision` is deterministic as if you ask to check out the same hash you will always get the same contents.
```
jglick: For mergeable PRs in `github-branch-source` it is actually a pair of commits—base branch and proposed branch. Worth mentioning that these things can be composite structures.

docs/implementation.adoc
jglick: IIUC AsciiDoc `--` renders as `—`.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+When a consuming plugin is processing a `jenkins.scm.api.SCMHeadEvent`, it will be necessary for the consuming plugin to revalidate the head instance against the `jenkins.scm.api.SCMSourceCriteria`.
+Even if we could trust the content of the `jenkins.scm.api.SCMHeadEvent`, the revalidation would be a necessary step as we have no way of determining from a `jenkins.scm.api.SCMHeadEvent` whether the change has:
+
+* Transitioned the criteria matching from `false` to `true`;
+* Maintained the criteria matching in its current state; or
+* Transitioned the criteria matching from `true` to `false`.
+
+Thus *every* SCM API consuming plugin that listens for a `jenkins.scm.api.SCMEvent` will need to validate that event *against the criteria*.
+This revalidation would be necessary even if we received a signed trusted event from the backing source control system.
```
jglick: Mm, unfortunate from a performance perspective.

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+
+Each `jenkins.scm.api.SCMSourceDescriptor` should provide the concrete instances of the `jenkins.scm.api.SCMHeadCategory` that are potentially generated by their `jenkins.scm.api.SCMSource` instances.
+Then each `jenkins.scm.api.SCMSource` instance can filter down that list to the actual categories that may be returned by that specific source.
+For example, a GitHub source may return "Branches", "Pull Requests" and "Tags" but the user may have configured their specific source for a specific project to only build "Branches" and "Tags".
+
+In an analgous way, the `jenkins.scm.api.SCMSource` instances themselves may have different terminology for each of the different source control systems:
+
+* GitHub uses the term "Repository" to refer to primary repositories
+* GitHub uses the term "Fork" to refer to forks of the primary repositories
+* Accurev uses the term "Depot" to refer to repositories (using the term "repository" to refer to the collection of "depots")
```
jglick: again CC @p4paul

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+        return true; // hopefully you do
+    }
+
+    @Override
+    public boolean requiresWorkspaceForPolling() {
+        return false; // hopefully you don't
+    }
+
+    // for easier interop with SCMSource
+    public void setSCMHead(@NonNull SCMHead head, @CheckForNull SCMRevision revision) {
```
jglick: Hey, when was this introduced? And why would you need it, when `SCMSource.build(SCMHead, SCMRevision)` does the same AFAICT?

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+                                                   @Nullable FilePath workspace, @Nonnull TaskListener listener,
+                                                   @Nonnull SCMRevisionState baseline)
+            throws IOException, InterruptedException {
+        if (baseline instanceof MySCMRevisionState) {
+            //
+            // get current revision in SCM
+            // if your implementation of requiresWorkspaceForPolling() returns true then the
+            // workspace and launcher parameters should be non-null and point to a
+            // workspace and node to use for the comparison
+            // NOTE: requiring a workspace for polling is a realy bad user experience
```
jglick: We should fix the Git plugin to use the local cache when path exclusions etc. are configured…

docs/implementation.adoc
KostyaSha: image

docs/implementation.adoc
```diff
@@ -0,0 +1,507 @@
...
+        this.revision = revision;
+    }
+
+    public MySCMRevision getRevision() {
+        return revision;
+    }
+}
+----
+
+The easiest format for the change log on disk is just to serialize the list of log entries using `XStream`.
```
jglick: Actual implementations tend to use a native command instead.

docs/implementation.adoc
KostyaSha: Yes, i'm monitoring this and would like to make multibranch support and implement non-workflow multibranch, but all this WIP APIs makes difficult understand what is happening.

docs/implementation.adoc
KostyaSha: And maybe some changelog types as third argument . i.e. follow merges and etc.

docs/implementation.adoc
stephenc: This is to make `SCMSource.build()` easy to implement

docs/implementation.adoc
stephenc: This is the introduction, I am giving an overview of everything that needs to be implemented... then I start going through the different parts one by one, in the order that makes the most sense... perhaps I just need to rename this section as Overview and put a short introduction ahead of the Overview

docs/implementation.adoc
KostyaSha: Why did you stick to GHOrganisation? Not sure whether i will use all this plugins, but I will be making the whole github representation in jenkins (like travis) based on oauth perms (this part already implemented). So it sounds like flat navigator.

docs/implementation.adoc
KostyaSha: I mean that changelog could be generate for diff in different ways - following parents that you so like to see in GH, so it not enough to ask about start-end, some kind or type of changelog should be also passed.

docs/implementation.adoc
KostyaSha: 👍 for asciidoc ;)

docs/implementation.adoc
stephenc: done

docs/implementation.adoc
stephenc: I think you can construct whatever change log you want by using repeated calls to `SCMFileSystem.of(SCMSource, SCMHead, SCMRevision)` followed by `SCMFileSystem.changesSince(SCMRevision, OutputStream)`.

To do anything more advanced would requiring enforcing a contract on the changelog entries which - while something that would be really nice to have - is not something I have the energy to try and drive through

docs/implementation.adoc
stephenc: I've rephrased this note to explain a bit better why I said more than 1 second

docs/implementation.adoc
stephenc: I have added some clarification and allowed implementations to declare some events as trusted.

The reality, however, is that trusted and untrusted events are not a concern of the consumer. The just allow for the implementations to optimize somewhat in certain cases.

docs/implementation.adoc
stephenc: Yep... but off topic here

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -253,13 +322,47 @@ public String toString() {
...
+         */
+        TRUSTED(true),
+        /**
+         * Signals that the {@link EventData} annotated method's return value or annotated field content is untrusted.
+         */
+        UNTRUSTED(false),
+        /**
+         * Signals that the {@link EventData} annotated method's return value or annotated field content's is derived
+         * from the event payload. To determine the trustability of the data for a specific event,
+         * consult {@link #isPayloadTrusted()}.
```
jglick: I am baffled here. `isPayloadTrusted` consults `Trustability.trusted`, which for `PAYLOAD` is always `false`.

src/main/java/jenkins/scm/api/SCMFile.java
```diff
@@ -40,6 +40,16 @@
  * @author Kohsuke Kawaguchi
  */
 public abstract class SCMFile {
+
+    /**
+     * Cache of the last modified timestamp, to allow repeated calls to minimize the number of network round trips.
+     */
+    private Long modified;
```
jglick: Should any caching not be the responsibility of the implementation? After all, it may have a better way of doing so. `GitSCMSource` in particular can do metadata lookups cheaply enough that there is no point in caching them. And even a remote-based source may have a better way of keeping a cache.

src/main/java/jenkins/scm/api/SCMHead.java
```diff
@@ -155,47 +146,34 @@ public String toString() {
...
-     * @see Actionable#getAllActions
+     * Returns an empty list.
+     * @return an empty list
      * @since 1.1
+     * @deprecated this was added to the API in error. Retained for backwards binary compatibility only. Use
+     * {@link SCMSource#fetchActions(SCMHead, SCMHeadEvent, TaskListener)} to get the actions associated with a
+     * {@link SCMHead}
      */
+    @Restricted(DoNotUse.class)
+    @Deprecated
```
jglick: A restriction on top of a deprecation is overkill.

src/main/java/jenkins/scm/api/SCMTriggerItemSCM.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.tools.ant.ExtensionPoint;
+
+/**
+ * Extension point to allow modern {@link SCM} implementations to auto-register call-back hooks for
+ * {@link SCMTriggerItem} instaces without having to implement a custom {@link Trigger}.
+ *
+ * @since 2.0
+ * @see SCMTriggerItemSCMNotifier
+ */
+public abstract class SCMTriggerItemSCM extends ExtensionPoint {
```
jglick: What an awful name. What does this mean?

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead.java
```diff
@@ -20,46 +20,40 @@
...
-    public ChangeRequestSCMHead(@NonNull String name) {
-        super(name);
-    }
+    @Exported
+    @NonNull
+    String getId();
 
     /**
-     * Returns the {@link ChangeRequestAction} of this {@link ChangeRequestSCMHead}.
+     * Branch to which this change would be merged or applied if it were accepted.
```
jglick: Is there any mechanism to signal that this has changed since the last call, for example in the case of a GH PR with a modified base, or can an implementation freely return a different value from call to call?

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead.java
```diff
@@ -20,46 +20,40 @@
...
- * @since FIXME
+ * @since 2.0
  */
-public abstract class ChangeRequestSCMHead extends SCMHead {
+public interface ChangeRequestSCMHead extends SCMHeadMixin {
+
     /**
-     * Constructor.
+     * Identifier of this change request.
+     * Expected to be unique among requests coming from a given {@link SCMSource}.
```
jglick: Is it also expected to be immutable?

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -47,6 +50,28 @@ protected SCMFileSystem(@CheckForNull SCMRevision rev) {
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void close() throws IOException {
```
jglick: (seems to have been resolved now)

src/main/java/jenkins/scm/api/SCMHeadEvent.java
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public boolean isMatch(@NonNull SCMSource source) {
+        return !heads(source).isEmpty();
+    }
+
+    /**
+     * Returns the {@link SCMHead} for the supplied {@link SCMSource} that this event corresponds to.
+     * <p>
+     * <strong>DO NOT TRUST THE RETURN VALUES.</strong> Data from events should only be used as a rumour that requires
+     * verification.
```
jglick: (resolved, I suppose)

src/main/java/jenkins/scm/api/SCMFile.java
stephenc: The modified cache may be overkill. The type cache I see as not being so. I have seen far far too many people do things like

```
File f = ...
if (file.exists() && file.isFile()) {
  ...
}
```

I'll remove the `modified` cache, especially as it is proving difficult for the important implementations of `SCMFileSystem` to be able to get the last modified timestamp of a file without a flurry of API calls.

src/main/java/jenkins/scm/api/SCMFile.java
```diff
@@ -42,22 +46,98 @@
...
+    protected SCMFile() {
+        this.parent = null;
+        this.name = "";
+    }
+
+    /**
+     * Constructor for any entry that is not the root.
+     *
+     * @param parent the parent reference or {@code null} if this is the root object.
+     * @since 2.0
```
jglick: Missing Javadoc for `name`; can it includes `/`s?

src/main/java/jenkins/scm/api/SCMHead.java
```diff
@@ -155,47 +146,34 @@ public String toString() {
...
     @CheckForNull
     public <T extends Action> T getAction(@NonNull Class<T> type) {
-        for (Action action : getAllActions()) {
-            if (type.isInstance(action)) {
-                return type.cast(action);
-            }
-        }
         return null;
```
jglick: At least maintain compatibility for callers of `ChangeRequestAction`: if `this instanceof ChangeRequestSCMHead`, return an action delegating to the new APIs. That should allow current `blueocean-pipeline-api-impl` releases to work.

src/main/java/jenkins/scm/api/SCMEvent.java
stephenc: So subclasses can override `payloadTrustability` for some payloads to declare that their specific payload was, e.g. signed.

There may be portions of the payload that are trusted and other portions that are not. Provider specific event sub-classes can use the `TRUSTED`, `PAYLOAD` and `UNTRUSTED` annotation values to clarify which parts are trusted, which parts come from the payload (which may or may not be trusted) and which parts are just plain untrusted.

This comes into play if you have an event delivery subsystem where you call back to the SCM to say "did you send this event" and the response is yes (so you know the head and revision but nothing else)

stephenc:
@jglick are were there yet?

src/main/java/jenkins/scm/api/SCMEvent.java
jglick: Is there some example of all this? In `github-branch-source`?

src/main/java/jenkins/scm/api/SCMFile.java
```diff
@@ -78,6 +78,7 @@ protected SCMFile() {
      * Constructor for any entry that is not the root.
      *
      * @param parent the parent reference or {@code null} if this is the root object.
+     * @param name   the name of this entry.
```
jglick: So is `/` illegal here? If so, a `child` implementation will be obliged to construct intermediate `SCMFile` directory objects, which seems onerous.

IMO rather than a `parent` field there should just be a backreference to the `SCMFileSystem` root, and a full `String path`. `getParent` can then simply construct a new flyweight object by stripping off the last path component.

src/main/java/jenkins/scm/api/SCMHead.java
```diff
@@ -207,7 +210,8 @@ public String toString() {
     @Deprecated
     @CheckForNull
     public <T extends Action> T getAction(@NonNull Class<T> type) {
-        return null;
+        // TODO should be deleted but old versions of blue ocean api rely on this
+        return ChangeRequestAction.class.equals(type) ? (T)new ChangeRequestAction(){} : null;
```
jglick: No, I meant to actually implement the methods!

src/main/java/jenkins/scm/api/SCMHeadMixinEqualityGenerator.java
```diff
@@ -0,0 +1,488 @@
...
+import static org.kohsuke.asm5.Opcodes.LLOAD;
+import static org.kohsuke.asm5.Opcodes.LSTORE;
+import static org.kohsuke.asm5.Opcodes.RETURN;
+
+/**
+ * Generates {@link SCMHeadMixin.Equality} instances for concrete {@link SCMHead} instance types.
+ * We need {@link SCMHead} instances to perform equality based on the {@link SCMHead#getName()} plus all the property
+ * values declared on the {@link SCMHeadMixin} interfaces implemented by the {@link SCMHead} concrete type.
+ * As {@link SCMHead#equals(Object)} is expected to be a hot method, we'd much rather avoid using reflection, so
+ * instead we use bytecode generation to create our {@link SCMHeadMixin.Equality} subclass for us.
```
jglick: This is crazy. Delete it and make `SCMHead.equals/hashCode` be `abstract` with the above contract.

src/main/java/jenkins/scm/api/SCMHead.java
stephenc: No. We are not implementing these methods. I would much rather that blue ocean fixes this asap rather than leave them with something that is built on hacks. There is no route from an SCMHead to the persisted actions, so all we can return here is a dummy... which was why I originally said _just return null_. Happy to go back to that, not happy to do any more than this

src/main/java/jenkins/scm/api/SCMFile.java
stephenc: My github implementation creates the intermediates without making remote calls, only making remote calls for those intermediates that get checked.

If you are really objecting I can move some of this stuff to a Base class in `jenkins.scm.impl` but essentially I see all implementations being simpler with this than the other way

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -155,21 +181,104 @@ protected final SCMSourceCriteria getCriteria() {
...
+            SCMSourceCriteria hopefullyNull = compatibilityHack.get();
+            compatibilityHack.set(criteria == null ? nullSCMSourceCriteria : criteria);
+            try {
+                retrieve(observer, listener);
+            } finally {
+                if (hopefullyNull != null) {
+                    // performance is going to be painful if you are nesting them
+                    compatibilityHack.set(hopefullyNull);
+                } else {
+                    compatibilityHack.remove();
```
stephenc: Not quite. `set(null)` will leave an entry in the `ThreadLocalMap` whereas `remove()` clears the entry.

src/test/java/jenkins/scm/impl/SingleSCMNavigatorTest.java
stephenc: We are testing the `getName()` method and ensuring that it always returned the supplied name. If we pass in a constant name, then an implementation of `getName()` that just happens to always return whatever constant name we passed in would also pass the test.

We could use two instantiations with different names as proof that the name is passed through, but it is simpler to just ensure that the name is different every time.

src/main/java/jenkins/scm/api/SCMEvent.java
stephenc: No examples yet. I suspect that "trusted" events are YANGNI except in the cases where the consumer is not using a criteria but you seemed keen on them. The point of these annotations is to expose to consumers that I have not anticipated about the trustability of the data from the events

src/main/java/jenkins/scm/api/SCMHead.java
stephenc: We use `@Restricted` to flag methods for use via stapler bindings. I have the `@Deprecated` to differentiate this is something we really do not want you using. I'd much rather just remove the method

src/main/java/jenkins/scm/api/SCMFile.java
```diff
@@ -41,73 +44,237 @@
...
-     * @return null if there's no file/directory at the path represented by it.
-     * @throws IOException if an error occurs while performing the operation.
+     * @return the pathname of this file.
+     * @since 2.0
+     */
+    @NonNull
+    public String getPath() {
+        if (parent == null) {
+            // root node
+            return "";
```
rsandell: You seem to be including the name below so shouldn't this be returning `getName()`?

src/main/java/jenkins/scm/api/SCMHeadMixinEqualityGenerator.java
rsandell: Well, I'm not sure but I think I'd prefer fixing any eventual problems in those plugins than trying to figure out a future bug in this code :)

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -67,10 +187,109 @@ public SCMNavigatorDescriptor getDescriptor() {
...
+     * {@link Item} for example by providing a {@link TransientActionFactory} implementation that reports these
+     * persisted actions separately (for example {@link AbstractProject#getActions()} returns an immutable list,
+     * so there is no way to persist the actions from this method against those sub-classes, instead the actions
+     * need to be persisted by some side mechanism and then injected into the {@link Actionable#getAllActions()}
+     * through a {@link TransientActionFactory} ignoring the cognitive dissonance triggered by adding non-transient
+     * actions through a transient action factory... think of it instead as a {@code TemporalActionFactory} that adds
+     * actions that can change over time)
+     *
+     * @param owner    the owner of this {@link SCMNavigator}.
+     * @param event    the (optional) even to use when fetching the actions. Where the implementation is
```
rsandell: even**t**

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -67,10 +187,109 @@ public SCMNavigatorDescriptor getDescriptor() {
...
+     *                 able to trust the event, it may use the event payload to reduce the number of
+     *                 network calls required to obtain the actions.
+     * @param listener the listener to report progress on.
+     * @return the map of {@link Action} instances to persist, keyed by the class of action. Keys with {@code null}
```
rsandell: The return type is not a `Map` so...?

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -67,10 +187,109 @@ public SCMNavigatorDescriptor getDescriptor() {
...
+    /**
+     * SPI for {@link #fetchActions(SCMNavigatorOwner, SCMNavigatorEvent, TaskListener)}. Fetches any actions that
+     * should be persisted for objects related to the specified owner.
+     *
+     * @param owner    the owner of this {@link SCMNavigator}.
+     * @param event    the (optional) even to use when fetching the actions. Where the implementation is
+     *                 able to trust the event, it may use the event payload to reduce the number of
+     *                 network calls required to obtain the actions.
+     * @param listener the listener to report progress on.
+     * @return the map of {@link Action} instances to persist, keyed by the class of action. Keys with {@code null}
```
rsandell: The return type is not a `Map` so...?

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -67,10 +187,109 @@ public SCMNavigatorDescriptor getDescriptor() {
...
+     * values indicate actions that should be removed if present.
+     * @throws IOException          if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since 2.0
+     */
+    @NonNull
+    public List<Action> retrieveActions(@NonNull SCMNavigatorOwner owner,
```
rsandell: I don't really see the difference between this and `fetchActions`? Maybe my speed reading isn't up to snuff?

stephenc:
@jglick @rsandell I think this is ready for final review. I have addressed the stuff Jesse wanted except for the "agree to differ" on philosophy for the `SCMHead.equals()` where Jesse is lodging a dance card for "The Dance Of I Told You So™" and I am playing the maintainer card to ACK and continue

KostyaSha:
You can merge and make alpha releases imho. Any hotfix for current version could be released from branch. 
Would all this cross-references influence on freestyle-multibranch-project? Probably i can pick it for maintaining if all this APIs will be suitable for me.

src/main/java/jenkins/scm/api/SCMHeadMixinEqualityGenerator.java
jglick: We agreed to disagree.

src/main/java/jenkins/scm/api/actions/ChangeRequestAction.java
```diff
@@ -45,7 +45,7 @@
  * title and links and {@link ChangeRequestSCMHead#getId()} and
  * {@link ChangeRequestSCMHead#getTarget()}
  */
-@Restricted(NoExternalUse.class) // TODO should be DoNotUse.class but old versions of blue ocean api rely
+@Restricted(DoNotUse.class) // TODO should be DoNotUse.class but old versions of blue ocean api rely
```
jglick: so remove the comment

src/main/java/jenkins/scm/api/SCMFile.java
```diff
@@ -41,73 +44,237 @@
...
+    protected SCMFile() {
+        this.parent = null;
+        this.name = "";
+    }
+
+    /**
+     * Constructor for any entry that is not the root.
+     *
+     * @param parent the parent reference or {@code null} if this is the root object.
+     * @param name   the name of this entry.
```
jglick: For some reason my previous comment got hidden, so: is this expected to be a single path component only (no `/`)? If so, that must be documented, and preferably enforced with `IllegalArgumentException`.

src/main/java/jenkins/scm/api/actions/ChangeRequestAction.java
```diff
@@ -26,98 +26,61 @@
...
- * Should also be restricted to short metadata which can be quickly retrieved during {@link SCMSource#retrieve(SCMHeadObserver, TaskListener)}.
- * @see SCMHead#getAllActions
+ * Class retained for binary compatibility only for instances upgrading and deserializing data from disk.
  * @since 1.1
+ * @deprecated use {@link ContributorMetadataAction} for the author information, {@link ObjectMetadataAction} for the
+ * title and links and {@link ChangeRequestSCMHead#getId()} and
+ * {@link ChangeRequestSCMHead#getTarget()}
  */
-@ExportedBean
+@Restricted(DoNotUse.class) // TODO should be DoNotUse.class but old versions of blue ocean api rely
```
jglick: So remove the comment.

src/main/java/jenkins/scm/api/SCMTriggerItemSCM.java
```diff
@@ -0,0 +1,67 @@
...
+import hudson.triggers.SCMTrigger;
+import hudson.triggers.Trigger;
+import jenkins.scm.impl.SCMTriggerItemSCMNotifier;
+import jenkins.scm.impl.SCMTriggerListener;
+import jenkins.triggers.SCMTriggerItem;
+import org.apache.tools.ant.ExtensionPoint;
+
+/**
+ * Extension point to allow modern {@link SCM} implementations to auto-register call-back hooks for
+ * {@link SCMTriggerItem} instaces without having to implement a custom {@link Trigger}.
```
jglick: typo

src/main/java/jenkins/scm/api/SCMTriggerItemSCM.java
jglick: Perhaps `SCMTriggerItemSaveListener`? At any rate, Javadoc could clarify that the purpose of this class is to allow an SCM plugin to register webhooks that will be processed via the new event system, without having to create a phony `Trigger` (one with no `run` override nor `cronTabSpec`) merely to intercept `start` (and `stop`, though currently we do not offer a replacement).

Do we offer any replacement for `getProjectActions`? For example like `GitHubWebHookPollingAction`?

Is there any downstream example of this extension point being implemented, demonstrating that it can fully replace the old idiom?

src/main/java/jenkins/scm/impl/SCMTriggerListener.java
```diff
@@ -0,0 +1,90 @@
...
+                        continue;
+                    }
+                    SCMTrigger trigger = scmTriggerItem.getSCMTrigger();
+                    if (trigger == null || trigger.isIgnorePostCommitHooks()) {
+                        // if it is not enabled for triggering => ignore it
+                        continue;
+                    }
+                    for (SCM scm : scmTriggerItem.getSCMs()) {
+                        if (!event.isMatch(scm)) {
+                            // only interested in SCMs that match the event
```
jglick: This condition looks backwards.

src/main/java/jenkins/scm/api/SCMNavigatorOwners.java
```diff
@@ -0,0 +1,144 @@
...
+    @Extension
+    @SuppressWarnings("unused")// instantiated by Jenkins
+    public static class JenkinsItemEnumerator extends Enumerator {
+
+        /**
+         * {@inheritDoc}
+         */
+        public Iterator<SCMNavigatorOwner> iterator() {
+            Jenkins j = Jenkins.getActiveInstance();
+            return j.getAllItems(SCMNavigatorOwner.class).iterator();
```
rsandell: This looks really costly in terms of performance especially if it is used every time an event is fired.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -155,25 +205,195 @@ protected final SCMSourceCriteria getCriteria() {
...
+    /**
+     * Fetches the latest heads and corresponding revisions. Implementers are free to cache intermediary results
+     * but the call must always check the validity of any intermediary caches.
+     *
+     * @param observer an optional observer of interim results.
+     * @param listener the task listener.
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     */
+    protected void retrieve(@NonNull SCMHeadObserver observer, @NonNull TaskListener listener)
```
rsandell: Still confused about the difference between `retrieve` and `fetch`.

rsandell: Especially since they seem to call each other back and forth without real distinction.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -355,6 +578,166 @@ public boolean isObserving() {
...
+    @NonNull
+    protected List<Action> retrieveActions(@NonNull SCMRevision revision,
+                                           @CheckForNull SCMHeadEvent event,
+                                           @NonNull TaskListener listener)
+            throws IOException, InterruptedException {
+        return Collections.emptyList();
+    }
+
+    /**
+     * SPI for {@link #fetchActions(SCMHead, SCMHeadEvent, TaskListener)}. Fetches any actions that should be persisted for objects
```
rsandell: _**S**chedule **p**erformance **i**ndex_ or _**S**erial **P**eripheral **I**nterface_ ?

src/test/java/jenkins/scm/api/SCMHeadObserverTest.java
stephenc: oops. added test

rsandell:
I'm overall :bee:, some confusions on some naming and a performance concern for usage in large enterprises where this type of feature would be highly sought for.

stephenc:
@KostyaSha  the plan will be to just cut a 2.0 release. This has been designed to work with older implementations. 

The 2.0 designator is because the SPI abstract methods to implement have been shifted so that if an implementation updates its dependency on SCM API, its build will be broken until the SPI methods have been adapted.

But older implementations compiled against the older SCM API should work transparently on the new SCM API... ignoring the `ChangeRequestAction` and actions being exposed from `SCMHead` mistake which is unwound (but in as compatible a way as possible)

KostyaSha:
@stephenc could you make alpha/beta releases for referenced plugins? It too hard jumping between PRs in all related plugins.

stephenc:
@KostyaSha the branch-api PR is backwards compatible for older implementations too, so I was not going to bother with alpha/beta for that...  

the github-org-folder PR is just a KILL KILL DIE DIE

for github branch source I can do an alpha / beta, but I don't see much utility there... ditto for bitbucket branch source.

What happens in workflow multibranch w.r.t. alpha/beta releases is up to Jesse

KostyaSha:
I'm interested in https://github.com/jenkinsci/freestyle-multibranch-plugin

stephenc:
@KostyaSha AFAIR there are no changes required other than the dependency bump but I'll double check

stephenc:
@KostyaSha so freestyle-multibranch-project needs a method that was removed from the initial commit of branch-api to be deleted and the dependency on branch-api updated... basically:

```
diff --git a/pom.xml b/pom.xml
index 3ae853c..ac5b333 100644
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
     <dependency>
       <artifactId>branch-api</artifactId>
-      <version>1.0</version>
+      <version>2.0-SNAPSHOT</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
diff --git a/src/main/java/org/cloudbees/jenkins/multibranch/freestyle/FreeStyleMultiBranchProject.java b/src/main/java/org/cloudbees/jenkins/multibranch/freestyle/FreeStyleMultiBranchProject.java
index 650de21..fc10685 100644
--- a/src/main/java/org/cloudbees/jenkins/multibranch/freestyle/FreeStyleMultiBranchProject.java
+++ b/src/main/java/org/cloudbees/jenkins/multibranch/freestyle/FreeStyleMultiBranchProject.java
@@ -145,21 +145,6 @@ public class FreeStyleMultiBranchProject extends
             return new FreeStyleMultiBranchProject(parent, name);
         }
 
-        /**
-         * {@inheritDoc}
-         */
-        @NonNull
-        @Override
-        public List<SCMDescriptor<?>> getSCMDescriptors() {
-            List<SCMDescriptor<?>> result = new ArrayList<SCMDescriptor<?>>(SCM.all());
-            for (Iterator<SCMDescriptor<?>> iterator = result.iterator(); iterator.hasNext(); ) {
-                SCMDescriptor<?> d = iterator.next();
-                if (NullSCM.class.equals(d.clazz)) {
-                    iterator.remove();
-                }
-            }
-            return result; // todo figure out filtering
-        }
     }
 
     /**
```

I'd need to double check on whether additional decorations are necessary (which will take place when I push to get the branch-api changes released), but AFAIR the decorations are all in SCM-API so branch api implementations do not to do anything with them.

Now if you were using SCM API to create a competing solution to branch api... then you might have greater interest in SCM APIs decorations... but AFAIK branch-api is currently the only consumer

stephenc:
@KostyaSha and the `getSCMDescriptors()` method deletion is not part of this PR but the result of https://github.com/jenkinsci/branch-api-plugin/commit/33369eed86f91e8f47400dad1768d17aff90fe8c

KostyaSha:
@stephenc it would be great if somebody create diagram between all this parts.

@stephenc is it possible to  refactor `freestyle-multibranch-project` to not bundle all classes in one, do you plan maintain it?

src/main/java/jenkins/scm/api/SCMNavigatorOwners.java
stephenc: Also prone to memory leaks. A lazy iterator should be just as efficient because you are walking item and we cannot know up front whether a specific event matches a specific target... we have to ask the event against each SCM / SCMSource / SCMNavigator whether it is a match (e.g. somebody may have a secondary system for events that the SCMSource implementation does not know about)

stephenc: Also events can match more than one job, so *if* we could have constructed IDs of sources we'd have a `Map<SourceId,List<Item>>` that we would need to maintain... same for `SCMHead` and `SCM` and `SCMNavigator`

src/main/java/jenkins/scm/api/SCMSource.java
rsandell: +1

rsandell:
:bee:

stephenc:
@KostyaSha if you want to take ownership of freestyle-multibranch-project, you are welcome. I created it as an example of how to implement a multibranch that was simpler than literate (because literate includes parts of matrix and promoted builds) and to show the issue with the stapler UI and ancestor assumed to be Job by lots of plugins.

I probably have a few changes to push to make it more modern (but none are strictly required to work with 2.0 branch api)

KostyaSha:
@stephenc i would like to try (atm running locally) and trying to read all this refactorings. In any case it not released.

jglick:
@KostyaSha note that `freestyle-multibranch-project` has never been released. Most people run `multi-branch-project-plugin` instead, which is fairly actively maintained.

src/main/java/jenkins/scm/api/MethodUtils.java
```diff
@@ -53,6 +53,7 @@ static boolean isAbstract(Class<?> clazz, String methodName, Class<?>... paramet
      * Checks if the  method defined on the base type with the given arguments
      * are overridden in the given derived type.
      */
+    // TODO replace with core utility method once JENKINS-40250 is available in base version of Jenkins
```
jglick: Probably suffices to update to 1.633+ (i.e., 1.642.x.) where JENKINS-30002 was fixed.

KostyaSha:
@jglick will look also, but i guess if i do multirabnch it should work both on freestyle and workflow?

src/main/java/jenkins/scm/api/SCMHeadEvent.java
```diff
@@ -170,47 +169,27 @@ public static void fireNow(@NonNull final SCMHeadEvent<?> event) {
...
-        public Dispatcher(SCMHeadEvent<?> event) {
-            this.event = event;
+        @Override
+        protected void log(SCMEventListener l, Throwable e) {
+            LogRecord lr = new LogRecord(Level.WARNING,
+                    "SCMEventListener.onSCMHeadEvent(SCMHeadEvent) {0} propagated an exception"
+            );
+            lr.setThrown(e);
+            lr.setParameters(new Object[]{l});
+            LOGGER.log(lr);
```
jglick: Simpler to write:

```java
LOGGER.log(Level.WARNING, "SCMEventListener.onSCMHeadEvent(SCMHeadEvent) " + l + " propagated an exception", e);
```

Since `INFO` and above are logged by default, there is no expected performance savings from this more verbose idiom.

src/main/java/jenkins/scm/api/SCMNavigatorEvent.java
```diff
@@ -91,48 +91,27 @@ public static void fireNow(@NonNull final SCMNavigatorEvent<?> event) {
...
-        public Dispatcher(SCMNavigatorEvent<?> event) {
-            this.event = event;
+        @Override
+        protected void log(SCMEventListener l, Throwable e) {
+            LogRecord lr = new LogRecord(Level.WARNING,
+                    "SCMEventListener.onSCMSourceEvent(SCMSourceEvent) {0} propagated an exception"
+            );
+            lr.setThrown(e);
+            lr.setParameters(new Object[]{l});
+            LOGGER.log(lr);
```
jglick: ditto

src/main/java/jenkins/scm/api/SCMSourceEvent.java
```diff
@@ -114,47 +114,27 @@ public static void fireNow(@NonNull final SCMSourceEvent<?> event) {
...
-        public Dispatcher(SCMSourceEvent<?> event) {
-            this.event = event;
+        @Override
+        protected void log(SCMEventListener l, Throwable e) {
+            LogRecord lr = new LogRecord(Level.WARNING,
+                    "SCMEventListener.onSCMSourceEvent(SCMSourceEvent) {0} propagated an exception"
+            );
+            lr.setThrown(e);
+            lr.setParameters(new Object[]{l});
+            LOGGER.log(lr);
```
jglick: ditto

src/main/java/jenkins/scm/api/MethodUtils.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return {@code true} if the method does not exist or is abstract.
+     */
+    static boolean isAbstract(Class<?> clazz, String methodName, Class<?>... parameterTypes) {
+        Method m = getMethodImpl(clazz, methodName, parameterTypes);
+        return m == null || Modifier.isAbstract(m.getModifiers());
+    }
+
+    /**
+     * Checks if the  method defined on the base type with the given arguments
+     * are overridden in the given derived type.
```
jglick: 1.633+ specifically

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead.java
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @return an ID of some kind, such as a pull request number (decimal) or a Gerrit change ID
+     */
+    @Exported
+    @NonNull
+    String getId();
+
+    /**
+     * Branch to which this change would be merged or applied if it were accepted.
```
jglick: I still do not see any compelling reason why this should be considered part of the identity of the head, when GitHub (the service, not our _current_ plugin) supports changing it on an existing PR.

src/main/java/jenkins/scm/api/SCMHeadMixinEqualityGenerator.java
jglick: Will add a final plea to replace all this with something like

```java
public SCMHead(String name) {
    if (/* implies we are in a subclass */this instanceof SCMHeadMixIn &&
        !Util.isOverridden(SCMHead.class, getClass(), "equals", Object.class)) {
        LOGGER.log(Level.WARNING, "possibly incorrect equals/hashCode in {0}", getClass());
    }
    this.name = name;
}
```

plus suitable Javadoc allowing an implementation to simply and clearly define its head equality semantics.

jglick:
> if i do multirabnch it should work both on freestyle and workflow?

Not sure what you mean by “do multibranch” in this context. `multi-branch-project-plugin` supports freestyle, Maven, and matrix projects. `workflow-multibranch` supports Pipeline. Complementing that, you need an `SCMSource` corresponding to a specific SCM.

KostyaSha:
@jglick i mean implement PR / Branch triggering for both job types at once. As i understand freestyle-multibranch will act the same as workflow-multibranch.

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead.java
stephenc: ACK. *if* SCMHead had been retained as a pure String wrapper that just adds usage type information (and perhaps tracking of origin based on the SCMSource) then there would be no reason. 

Once GitHub and BitBucket started storing the Target in the SCMHead (and not in the SCMRevision) then we need this.

Keep in mind that the SCMRevision equality is based on the revisions's head, so if the Target changes, with this equals we get the revision always being changed and getting a rebuild... without it, for revisions that are just tracking the fork (but untrusted and using the Jenkinsfile from the target) will not get their rebuild... we could fix that by moving the target to SCMRevision, but the ship has sailed due to the usage in those two plugins and I see some additional utility in making the target be part of the head identity

stephenc:
@kostyasha if you implement the scm api for events fully then everything should all work through that and we can remove some of the horrible hacks that abound left, right and centre.

Branch-API job types should just trigger off of scm events. Old style standalone jobs using SCM directly will work if you integrate them with the new API. For some use cases you may need to either extend SCM2 or roll your own equivalency

jglick:
> Once GitHub and BitBucket started storing the Target in the `SCMHead` (and not in the `SCMRevision`)

I cannot speak to BitBucket. In the case of GitHub, the base branch is considered when calculating the `SCMRevision`.

> the `SCMRevision` equality is based on the revisions's head, so if the Target changes, with this `equals` we get the revision always being changed and getting a rebuild

Even without `SCMHead.equals` considering `target`, you will get a rebuild if the base branch is changed even without any change to the PR head.

> for revisions that are just tracking the fork

Non-`merge`?

> (but untrusted and using the `Jenkinsfile` from the target) will not get their rebuild

It will, because in that case after the rebase `PullRequestSCMRevision.baseHash` will point to something different, so the new revision will not be `equals` to that of the last build.

PR #19, opened 2016-12-15
SCMNavigator needs an ID of the thing being navigated

stephenc:
- Otherwise we cannot track the fetched actions across configuration tweaks

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -107,6 +107,13 @@ public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator>
                 // and even if we glom some legitimate text, the chances of a duplicate that is otherwise
                 // the same but has a critical difference in a regular value that happens to contain
                 // plugin="xys@abc" is sufficiently low.
+                //
+                // We can use a generated ID based on the initial serialization digest because any implementation
+                // that does not have an id() override will not have a fetchActions() override either
+                // so there will be no actions to retain anyway.
+                //
+                // Real implementations should not use this type of scheme for generating their id, rather
+                // they should construct their id explicitly.
                 id = getClass().getName() + "::" + Util.getDigestOf(Items.XSTREAM.toXML(this)
                         .replaceAll(" plugin=(('[^']+@[^']+')|(\"[^\"]+@[^\"]+\"))", ""));
             } else {
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -33,6 +33,7 @@ import hudson.model.AbstractProject;
 import hudson.model.Action;
 import hudson.model.Actionable;
 import hudson.model.Item;
+import hudson.model.Items;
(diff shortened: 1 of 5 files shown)


src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -64,12 +66,92 @@
             = new AlternativeUiTextProvider.Message<SCMNavigator>();
 
     /**
+     * Cache of the ID of this {@link SCMNavigator}.
+     * @since FIXME
```
jglick: No need for `@since` on a non-API field.

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -64,12 +66,92 @@
...
+        if (id == null) {
+            if (MethodUtils.isAbstract(getClass(), "id")) {
+                // we need to ensure that upgrading the plugin does not change the digest
+                // processing XML with a regex is usually a bad thing. In this case we have a sufficiently
+                // strict set of requirements to match (shortname)@(version) and the constraints on the
+                // shortname and version are such that we can have a high degree of certainty of a match
+                // and even if we glom some legitimate text, the chances of a duplicate that is otherwise
+                // the same but has a critical difference in a regular value that happens to contain
+                // plugin="xys@abc" is sufficiently low.
+                id = getClass().getName() + "::" + Util.getDigestOf(Items.XSTREAM.toXML(this)
```
jglick: Does the XStream form not already include this class name?

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -64,12 +66,92 @@
...
+            } else {
+                id = getClass().getName() + "::" + id();
+            }
+        }
+        return id;
+    }
+
+    /**
+     * If implementations are using {@link DataBoundSetter} on fields that affect the {@link #id()} calculation then
+     * those fields must call {@link #resetId()} if they may have invalidated the cached {@link #getId()}.
```
jglick: IMO better to leave the choice of whether a cache is needed up to the implementation, and just have

```java
public String getId() {
    return getClass().getName() + "::" + Util.getDigestOf(Items.XSTREAM.toXML(this)
        .replaceAll(" plugin=(('[^']+@[^']+')|(\"[^\"]+@[^\"]+\"))", ""));
}
```

After all, the current system will not correctly `resetId` when there are configuration changes in implementations failing to override `id`.

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: Class name is for debugabilty and to namespace impls

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: Needed on SPI fields such as this

src/main/java/jenkins/scm/api/SCMNavigator.java
stephenc: Old impls will not have `fetchActions` so probably a moot point for them!

PR #22, opened 2017-01-05
[FIXED JENKINS-40828] Provide a way for tests using MockSCMController to inject failures

stephenc:
See [JENKINS-40828](https://issues.jenkins-ci.org/browse/JENKINS-40828)

@reviewbybees

### DIFF
--- a/src/test/java/jenkins/scm/impl/mock/MockFailure.java
+++ b/src/test/java/jenkins/scm/impl/mock/MockFailure.java
@@ -1,0 +1,46 @@
+/*
+ * The MIT License
+ *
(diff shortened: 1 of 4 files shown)


src/test/java/jenkins/scm/impl/mock/MockFailure.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 CloudBees, Inc.
```
stephenc: That code was actually written by me last year...

rsandell:
:bee:

PR #24, opened 2017-01-20
[JENKINS-41121] Provide an API for SCMHead migration

stephenc:
See [JENKINS-41121](https://issues.jenkins-ci.org/browse/JENKINS-41121) requires matching changes in branch-api and github-branch-source plugins (bitbucket-branch-source is not affected as it already had done the right thing and implemented its own `SCMHead` (though did not store enough information for PRs so all PRs will rebuild and there is nothing we can do as the target branch was not retained)

@reviewbybees

(i'll be linking the downstream PRs to this)

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMHeadMigration.java
+++ b/src/main/java/jenkins/scm/api/SCMHeadMigration.java
@@ -1,3 +1,26 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017, CloudBees, Inc..
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
 package jenkins.scm.api;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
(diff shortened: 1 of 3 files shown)


src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -0,0 +1,126 @@
+package jenkins.scm.api;
+
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+/**
+ * If a {@link SCMSource} plugin needs to migrate the implementation classes for {@link SCMHead} this extension
+ * point allows the plugin to register type migrations. For speed of migration implementations should just jump direct
```
jglick: directly

src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -0,0 +1,126 @@
...
+    private final Class<H> headClass;
+    /**
+     * The {@link SCMRevision} that the migration applies to.
+     */
+    private final Class<R> revisionClass;
+
+    /**
+     * Constructor.
+     *  @param sourceClass The {@link SCMSource} that the migration applies to.
+     * @param headClass   The {@link SCMHead} that the migration applies to.
```
jglick: odd whitespaces

src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -0,0 +1,126 @@
...
+     * @param revisionClass
```
jglick: delete if empty

src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -0,0 +1,126 @@
...
+    protected SCMHeadMigration(Class<S> sourceClass, Class<H> headClass, Class<R> revisionClass) {
+        this.sourceClass = sourceClass;
+        this.headClass = headClass;
+        this.revisionClass = revisionClass;
+    }
+
+    /**
+     * Gets the {@link SCMHead} that the migration applies to.
+     *
+     * @return the {@link SCMHead} that the migration applies to.
```
jglick: `SCMSource` I think you meant.

src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Perform a migration.
+     * @param source the source instance.
+     * @param head the candidate head.
+     * @return the migrated head or {@code null} if the migration was not appropriate.
+     */
+    @CheckForNull
+    public abstract SCMHead migrate(@NonNull S source, @NonNull H head);
```
jglick: Why is there not a default `null` return value?

src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -0,0 +1,126 @@
...
+
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+/**
+ * If a {@link SCMSource} plugin needs to migrate the implementation classes for {@link SCMHead} this extension
+ * point allows the plugin to register type migrations. For speed of migration implementations should just jump direct
+ * to the final end-point and not expect recursive chain walking.
```
jglick: Seems like a premature optimization.

stephenc: well it also prevents having to worry about somebody accidentally setting up a circular migration chain via extensions

src/main/java/jenkins/scm/api/SCMHeadMigration.java
stephenc: Because this is the critical method I expect you to implement

PR #25, opened 2017-01-25
[JENKINS-41453] Migration code was overly zealous

stephenc:
See [JENKINS-41453](https://issues.jenkins-ci.org/browse/JENKINS-41453) 

A more generic fix than https://github.com/jenkinsci/github-branch-source-plugin/pull/111

@reviewbybees esp @michaelneale


src/main/java/jenkins/scm/api/SCMHeadMigration.java
```diff
@@ -124,8 +124,8 @@ public SCMRevision migrate(@NonNull S source, @NonNull R revision) {
     @NonNull
     public static SCMHead readResolveSCMHead(@NonNull SCMSource source, @NonNull SCMHead head) {
         for (SCMHeadMigration m : ExtensionList.lookup(SCMHeadMigration.class)) {
-            if (m.sourceClass.isInstance(source)
-                    && m.headClass.isInstance(head)) {
+            if (m.sourceClass.equals(source.getClass())
```
jglick: Can use `==` for `Class`es.

svanoort:
🐝  It's a precise and limited-scope migration that doesn't touch potentially unsafe subclasses.  Nice.

PR #27, opened 2017-02-07
[FIXED JENKINS-41795] Add field to track event origins

stephenc:
See [JENKINS-41795](https://issues.jenkins-ci.org/browse/JENKINS-41795)

@reviewbybees

### DIFF
--- a/docs/implementation.adoc
+++ b/docs/implementation.adoc
@@ -1279,9 +1279,9 @@ The webhook receiver would start by parsing the payload and then create the appr
 JsonNode json = new ObjectMapper().readTree(payload);
 String event = json.path("event").asString();
 if ("branch-create".equals(event)) {
-    SCMHeadEvent.fireNow(new MyBrachSCMHeadEvent(Type.CREATED, json);
+    SCMHeadEvent.fireNow(new MyBranchSCMHeadEvent(Type.CREATED, json);
 } else if ("branch-update".equals(event)) {
-    SCMHeadEvent.fireNow(new MyBrachSCMHeadEvent(Type.UPDATED, json);
+    SCMHeadEvent.fireNow(new MyBranchSCMHeadEvent(Type.UPDATED, json);
 } // else etc
 ----
 
@@ -1289,9 +1289,9 @@ Because each event from the source control system has a 1:1 correspondance with 
 
 [source,java]
 ----
-public class MyBrachSCMHeadEvent extends SCMHeadEvent<JsonNode> {
+public class MyBranchSCMHeadEvent extends SCMHeadEvent<JsonNode> {
 
-    public MyBrachSCMHeadEvent(@NonNull Type type, JsonNode payload, String origin) {
+    public MyBranchSCMHeadEvent(@NonNull Type type, JsonNode payload, String origin) {
         super(type, payload, origin);
     }
 
--- a/docs/implementation.adoc
+++ b/docs/implementation.adoc
@@ -1163,13 +1163,13 @@ public class MySCMWebHook implements UnprotectedRootAction {
         for (event : payload) {
             switch (eventType) {
                 case HEAD:
-                    SCMHeadEvent.fireNow(new MySCMHeadEvent(eventType, payload);
+                    SCMHeadEvent.fireNow(new MySCMHeadEvent(eventType, payload, SCMEvent.originOf(req));
                     break;
                 case SOURCE:
-                    SCMHeadEvent.fireNow(new MySCMSourceEvent(eventType, payload);
+                    SCMHeadEvent.fireNow(new MySCMSourceEvent(eventType, payload, SCMEvent.originOf(req));
                     break;
                 case NAVIGATOR:
-                    SCMHeadEvent.fireNow(new MySCMNavigatorEvent(eventType, payload);
+                    SCMHeadEvent.fireNow(new MySCMNavigatorEvent(eventType, payload, SCMEvent.originOf(req));
                     break;
             }
         }
@@ -1291,8 +1291,8 @@ Because each event from the source control system has a 1:1 correspondance with 
 ----
 public class MyBrachSCMHeadEvent extends SCMHeadEvent<JsonNode> {
 
-    public MyBrachSCMHeadEvent(@NonNull Type type, JsonNode payload) {
-        super(type, payload);
+    public MyBrachSCMHeadEvent(@NonNull Type type, JsonNode payload, String origin) {
+        super(type, payload, origin);
     }
 
     @Override
--- a/src/main/java/jenkins/scm/api/SCMEvent.java
+++ b/src/main/java/jenkins/scm/api/SCMEvent.java
@@ -25,6 +25,7 @@
 
 package jenkins.scm.api;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionList;
 import hudson.model.Cause;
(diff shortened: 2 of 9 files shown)


docs/implementation.adoc
```diff
@@ -1291,8 +1291,8 @@ Because each event from the source control system has a 1:1 correspondance with
 ----
 public class MyBrachSCMHeadEvent extends SCMHeadEvent<JsonNode> {
 
-    public MyBrachSCMHeadEvent(@NonNull Type type, JsonNode payload) {
-        super(type, payload);
+    public MyBrachSCMHeadEvent(@NonNull Type type, JsonNode payload, String origin) {
```
jglick: Typo BTW.

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -110,16 +117,40 @@
...
+        this(type, timestamp, payload, originOf(Stapler.getCurrentRequest()));
+    }
+
+    /**
+     * Constructor to use when the timestamp is available from the external SCM.
+     *
+     * @param type      the type of event.
+     * @param timestamp the timestamp from the external SCM (see {@link System#currentTimeMillis()} for start and units)
+     * @param payload   the original provider specific payload.
+     * @param origin    the (optional) origin of the event, e.g. a hostname, etc
```
jglick: `etc.`

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -194,6 +240,17 @@ public P getPayload() {
     }
 
     /**
+     * Gets the origin of the event, e.g. a hostname, etc.
+     *
+     * @return the origin of the event (or {@code "?"} if the origin is unknown)
```
jglick: Better to make `ORIGIN_UNKNOWN` a `public` constant.

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -247,14 +304,79 @@ public int hashCode() {
...
     }
 
     /**
+     * Helper method to get the origin of an event from a {@link StaplerRequest}.
+     *
+     * @param req the {@link StaplerRequest}.
+     * @return the origin of the event.
+     */
+    @CheckForNull
+    public static String originOf(@CheckForNull StaplerRequest req) {
```
jglick: Can you use `HttpServletRequest`?

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -247,14 +304,79 @@ public int hashCode() {
...
+                payload,
+                StringUtils.defaultIfBlank(origin, ORIGIN_UNKNOWN)
         );
     }
 
     /**
+     * Helper method to get the origin of an event from a {@link StaplerRequest}.
+     *
+     * @param req the {@link StaplerRequest}.
+     * @return the origin of the event.
```
jglick: Be more specific about the format, and/or give an example. And under which conditions might it be null?

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -247,14 +304,79 @@ public int hashCode() {
...
+                }
+                last = StringUtils.trim(remote);
+                result.append(last);
+            }
+        }
+        String remoteHost = req.getRemoteHost();
+        String remoteAddr = req.getRemoteAddr();
+        if (last == null || (!(StringUtils.equals(last, remoteHost) || StringUtils.equals(last, remoteAddr)))) {
+            if (last != null) {
+                result.append(" => ");
```
jglick: or `→`

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -247,14 +304,79 @@ public int hashCode() {
...
+                port = Integer.parseInt(portStr);
+            } catch (NumberFormatException e) {
+                port = req.getRemotePort();
+            }
+        } else {
+            port = req.getRemotePort();
+        }
+        if (!("http".equals(scheme) && port == 80 || "https".equals(scheme) && port == 443)) {
+            result.append(':');
+            result.append(port);
```
jglick: Seems rather complicated. Is testing with a mock feasible here?

docs/implementation.adoc
```diff
@@ -1279,20 +1279,20 @@ The webhook receiver would start by parsing the payload and then create the appr
...
+    SCMHeadEvent.fireNow(new MyBranchSCMHeadEvent(Type.UPDATED, json);
 } // else etc
 ----
 
 Because each event from the source control system has a 1:1 correspondance with the events in the SCM API the implementation of each event can be fairly straightforward:
 
 [source,java]
 ----
-public class MyBrachSCMHeadEvent extends SCMHeadEvent<JsonNode> {
+public class MyBranchSCMHeadEvent extends SCMHeadEvent<JsonNode> {
```
svanoort: Isn't this breaking binary compatibility?  

Maybe more of a 🐜  than a 🐛  if these APIs are not published yet, but be careful here to make sure this is updated wherever needed.

src/main/java/jenkins/scm/api/SCMEvent.java
svanoort: Concur with Jesse on both points.  I can follow what you're doing here but would seriously look at the Apache and Guava dependencies to see if we have a utility that can cut out some of this nastiness.

docs/implementation.adoc
stephenc: This is documentation

svanoort: NVM you're right, I misread the diff

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -247,14 +309,83 @@ public int hashCode() {
...
+    @CheckForNull
+    public static String originOf(@CheckForNull HttpServletRequest req) {
+        if (req == null) {
+            return null;
+        }
+        String last = null;
+        StringBuilder result = new StringBuilder();
+        // TODO RFC 7239 support once standard is approved
+        String header = req.getHeader("X-Forwarded-For");
+        if (StringUtils.isNotBlank(header)) {
```
svanoort: I think this can be replaced by a utility function from Guava, which has a StringJoiner that should do it as a one-liner.

svanoort: Specifically the string joining with separator (and null handling)

stephenc: it could be replaced by a utility from StringUtils *except* we need to catch the last value in case it matches the remote host/addr

src/main/java/jenkins/scm/api/SCMEvent.java
svanoort: I wonder if we don't have an API somewhere in core that will generate a user-readable string from the forwarding headers, for proxy?  @stephenc have you looked to see?

src/main/java/jenkins/scm/api/SCMEvent.java
stephenc: if we don't catch the last value then you can end up with a forwarding listed twice (though I would argue that such a case is a badly configured forwarding gateway, we should be magnanimous in what we accept and tidy it up!)

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -247,14 +309,83 @@ public int hashCode() {
...
                 type,
                 timestamp,
-                payload
+                payload,
+                StringUtils.defaultIfBlank(origin, ORIGIN_UNKNOWN)
         );
     }
 
     /**
+     * Helper method to get the origin of an event from a {@link HttpServletRequest}. The current format is the
```
KostyaSha: If it helper and not private, why not put it into helper class? Scrolling class with helpers and nested classes makes eyes red. Though https://github.com/jenkinsci/ldap-plugin/blob/master/src/main/java/hudson/security/LDAPSecurityRealm.java is still record with 15 classes.

PR #30, opened 2017-02-15
[JENKINS-42056] Fix PCT for 2.32.2

varyvol:
[JENKINS-42056](https://issues.jenkins-ci.org/browse/JENKINS-42056)

@reviewbybees @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,12 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>junit</artifactId>
+      <version>1.6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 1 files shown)


varyvol:
@jglick so I was looking at this one specifically: [link](https://github.com/jenkinsci/workflow-api-plugin/blob/master/Jenkinsfile#L1)

If I understand it correctly, that will make the plugin to be built against its baseline version and 2.32.2 version and, if any of them fails, the build fails?

How do you plan to maintain this for each Jenkins LTS version?

jglick:
> that will make the plugin to be built against its baseline version and 2.32.2 version and, if any of them fails, the build fails?

Correct. 2.32.2 is a bit arbitrary; what is more important is that it is something at least 2.19.x, so that test runs enforce plugin dependencies, which does not happen in the current 1.642.x baseline.

> How do you plan to maintain this for each Jenkins LTS version?

As above, there is no pressing need to update that version.

PR #31, opened 2017-02-17
[FIXED JENKINS-42150] Add latency options to mock scm impl in SCM API's test artifact

stephenc:
See [JENKINS-42150](https://issues.jenkins-ci.org/browse/JENKINS-42150)

@reviewbybees


src/test/java/jenkins/scm/impl/mock/MockSCMController.java
```diff
@@ -80,6 +82,16 @@ public static MockSCMController create() {
         return c;
     }
 
+    public MockSCMController withLatency(@NonNull MockLatency latency) {
+        this.latency = latency;
+        return this;
+    }
+
+    public MockSCMController withoutLatency() {
```
jglick: Just create a `MockLatency.none()` to cancel. That could also avoid null checks elsewhere.

PR #33, opened 2017-03-07
[FIXED JENKINS-42542] SCMHeadObserver.observe(SCMHead,SCMRevision) should be allowed to throw IO and Interrupted exceptions

stephenc:
See [JENKINS-42542](https://issues.jenkins-ci.org/browse/JENKINS-42542)

This is strictly speaking a breaking method signature change, the following is the rationale as to why we can make the change with minimal risk:

* _consumers_ of SCM API will be implementing their own subclasses of `SCMHeadObserver`, they will not (outside of unit tests) be calling `SCMHeadObserver.observe()` thus the widening of the method signature will not affect consumers plugin implementations. (though consumers may have test compilation errors when they upgrade their dependency)

* `SCMSource.fetch(...)` (and it's matching SPI, `SCMSource.retrieve(...)`) already declare throwing `IOException` and `InterruptedException` so _implementations_ of SCM API should not be affected by the signature change. At worst implementations will pick up a compiler error when they update their SCM API dependency - though a quick test revealed that git, mercurial, github and Bitbucket SCMSource implementations were all unaffected by the change.

* Running a newer SCM API with an implementation that were compiled against the older SCM API will not cause issues as:

    * Checked exceptions are not enforced at runtime, only at compile time.

    * The calls to `SCMHeadObserver.observe(...)` will all be within `SCMSource.retrieve(...)` which already propagates the same exceptions so consumers will be handling the exceptions anyway.

@reviewbybees

### DIFF
--- a/src/test/java/jenkins/scm/api/SCMEventTest.java
+++ b/src/test/java/jenkins/scm/api/SCMEventTest.java
@@ -209,7 +209,7 @@ public class SCMEventTest {
         Mockito.when(req.getHeader("X-Forwarded-Port")).thenReturn("443");
         Mockito.when(req.getHeader("X-Forwarded-For")).thenReturn("scm.example.com, gateway.example.com, proxy.example.com");
         Mockito.when(req.getRequestURI()).thenReturn("/jenkins/notify");
-        Mockito.when(req.getLocalPort()).thenReturn(8080);
+        Mockito.when(req.getRemotePort()).thenReturn(8080);
         Mockito.when(req.getRemoteHost()).thenReturn(null);
         Mockito.when(req.getRemoteAddr()).thenReturn("203.0.113.1");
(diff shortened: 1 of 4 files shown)


src/test/java/jenkins/scm/api/SCMEventTest.java
```diff
@@ -209,7 +209,7 @@ public void originOfForwardedRequestMultiHop() throws Exception {
         Mockito.when(req.getHeader("X-Forwarded-Port")).thenReturn("443");
         Mockito.when(req.getHeader("X-Forwarded-For")).thenReturn("scm.example.com, gateway.example.com, proxy.example.com");
         Mockito.when(req.getRequestURI()).thenReturn("/jenkins/notify");
-        Mockito.when(req.getLocalPort()).thenReturn(8080);
+        Mockito.when(req.getRemotePort()).thenReturn(8080);
```
daniel-beck: Unclear

stephenc: That shouldnt be part of this PR, but it is correct. Evaluate the origin of the event was reporting the remote port rather than the local port... only not caught because the tests were using mocks and made the same mistake

PR #34, opened 2017-03-13
[JENKINS-41736] Add ability for SCMEvents to describe themselves

stephenc:
See [JENKINS-41736](https://issues.jenkins-ci.org/browse/JENKINS-41736)

@reviewbybees esp @i386

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMEvent.java
+++ b/src/main/java/jenkins/scm/api/SCMEvent.java
@@ -43,6 +43,7 @@ import jenkins.util.Timer;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
 import org.apache.commons.lang.StringUtils;
+import org.jvnet.localizer.Localizable;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.Stapler;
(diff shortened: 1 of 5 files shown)


KostyaSha:
Why PR didn't get review?

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -256,6 +257,17 @@ public String getOrigin() {
     }
 
     /**
+     * Return a description of the event.
+     *
+     * @return the description or {@code null} if no description can be provided.
+     * @since TODO
+     */
+    @CheckForNull
+    public String description() {
```
KostyaSha: `getDescription()` to not confuse people with builders

stephenc: Well the other ones are not pure getters, e.g. `descriptionFor(SCMNavigator n)`

stephenc:
> Why PR didn't get review?

not needed, was done in personal time, I only mentioned the reviews as a courtesy... after 3 days...

src/main/java/jenkins/scm/api/SCMEvent.java
KostyaSha: Without verb before it very unclear what this method do, either getting, either setting, either reseting

KostyaSha:
> not needed, was done in personal time, I only mentioned the reviews as a courtesy... after 3 days...

I don't agree with such statement. scm-api is dependency for everybody now, so using veto because this plugin get into dependency while it had only 5-6 classes ( i check first releases added to git plugin) is not acceptable. Literally it installed in all instances https://wiki.jenkins-ci.org/display/JENKINS/SCM+API+Plugin 
Either github-plugin/git-plugin should have it optionally so other companies can exclude it from dependency tree.

stephenc:
@KostyaSha this is not a breaking change. It does not affect compatibility. It is an opt-in way for events to provide more details. Plugins do not need to update their SCM-API version to implement this (they can just pre-implement the overrides)

I thought you were asking about my employer's policy on code reviews, which I replied was not needed for this specific change.

PR #35, opened 2017-04-07
[JENKINS-43433] Allow SCMSource implementations to expose merge and origin of change request heads

stephenc:
See [JENKINS-43433](https://issues.jenkins-ci.org/browse/JENKINS-43433)

Todo:

- [x] Update Mock SCM provider to allow for this
- [x] Sanity check that API changes work with GitHub and Bitbucket without causing rebuild storms


stephenc:
@reviewbybees

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMRevision.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    private final SCMRevision target;
+
+    /**
+     * Constructor.
+     *
+     * @param head   the {@link SCMHead} that the {@link SCMRevision} belongs to.
+     * @param target the {@link SCMRevision} of the {@link ChangeRequestSCMHead#getTarget()}.
+     */
+    protected ChangeRequestSCMRevision(@NonNull H head, @NonNull SCMRevision target, boolean merge) {
```
rsandell: `merge` is unused

stephenc:
@rsandell I'm more thinking are the names and approach in general right.

This is being slightly driven by the way the GitHub Branch Source has implemented things... which could be the tail wagging the dog (at least for the mergeable mixin)

stephenc:
@rsandell we should be good to go on this once we get a re-review

src/main/java/jenkins/scm/api/SCMHeadOrigin.java
```diff
@@ -0,0 +1,195 @@
...
+     * A named fork. The name should be simple not an URL.
+     */
+    public static class Fork extends SCMHeadOrigin {
+        /**
+         * Standardize serialization.
+         */
+        private static final long serialVersionUID = 1L;
+
+        /**
+         * The name of the fork. This should
```
jglick: Should what?

src/main/java/jenkins/scm/api/mixin/ChangeRequestCheckoutStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+     * (perhaps with de-duplication).
+     * <p>
+     * In the event that the rebase operation cannot be completed then the checkout operation will fail.
+     * </p>
+     * <p>
+     * It is assumed that the algorithm for rebasing is deterministic given the revision of the change request and
+     * the revision of the {@link ChangeRequestSCMHead#getTarget()}
+     * </p>
+     */
+    REBASE,
```
jglick: Why does this need to be defined in the API? IOW why would a client need to differentiation this from `MERGE`? Obviously an implementation could perform any of a variety of merge-like strategies (squashes, for example).

PR #36, opened 2017-05-04
[JENKINS-43507] Allow SCMSource and SCMNavigator subtypes to share common traits

stephenc:
See [JENKINS-43507](https://issues.jenkins-ci.org/browse/JENKINS-43507)

@reviewbybees


src/main/java/jenkins/scm/api/trait/SCMBuilder.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Builder for a {@link SCM} instance. Typically instantiated in {@link SCMSource#build(SCMHead, SCMRevision)} or
+ * {@link SCMSource#build(SCMHead)} and then decorated by {@link SCMSourceTrait#applyToBuilder(SCMBuilder)} before
+ * calling {@link #build()} to generate the return value. Conventions:
+ * <ul>
+ * <li>The builder is not designed to be shared by multiple threads.</li>
+ * <li>All methods should be either {@code final} or {@code abstract} unless there is a documented reason for
+ * allowing overrides</li>
+ * <li>All "setter" methods will return {@link B} and be called "withXxx"</li>
+ * <li>All "getter" methods will called "xxx()" callers should not assume that the returned value is resistent
```
rsandell: will **be** called

src/main/java/jenkins/scm/impl/trait/WildcardSCMHeadFilterTrait.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Returns the pattern corresponding to the branches containing wildcards.
+     *
+     * @param branches the names of branches to create a pattern for
+     * @return pattern corresponding to the branches containing wildcards
+     */
+    private String getPattern(String branches) {
+        StringBuilder quotedBranches = new StringBuilder();
+        for (String wildcard : branches.split(" ")) {
```
rsandell: Maybe this is copied from somewhere to keep bahaviour from previous versions, otherwise there are nice utility methods in ANT that enables `**` and `?` style wildcards as well.

rsandell: https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/BranchConditional.java#L70

src/main/resources/jenkins/scm/api/form/traits.jelly
```diff
@@ -0,0 +1,128 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
```
rsandell: Would be nice if there was some comments inserted where this code deviates from a normal hetero-list.

src/test/java/jenkins/scm/impl/SingleSCMSourceTest.java
```diff
@@ -61,88 +70,125 @@
...
-        SCMSourceBuilder builder = new SCMSourceBuilder(source);
-        r.assertEqualDataBoundBeans(builder, r.configRoundtrip(builder));
+        MockSCMController c = MockSCMController.create();
+        try {
+            c.createRepository("foo");
+            SingleSCMSource source =
+                    new SingleSCMSource("the-id", "the-name", new MockSCM(c, "foo",
+                            new MockSCMHead("master"), null));
+            SCMSourceBuilder builder = new SCMSourceBuilder(source);
+            r.assertEqualDataBoundBeans(builder, r.configRoundtrip(builder));
```
rsandell: Does this even work? From what I've seen `configRoundtrip` normally just returns the parameter as is. Which either means you compare between old and old data, or new and new data.

rsandell:
fd9652e to 9fb3797 looks ok

src/main/java/jenkins/scm/api/trait/SCMSourceRequest.java
```diff
@@ -0,0 +1,553 @@
...
+        if (Thread.interrupted()) {
+            throw new InterruptedException();
+        }
+        if (isExcluded(head)) {
+            // not included
+            return !observer.isObserving();
+        }
+        I intermediate = intermediateFactory == null ? null : intermediateFactory.create();
+        if (!criteria.isEmpty()) {
+            SCMSourceCriteria.Probe probe = probeFactory.create(head, intermediate);
```
stephenc: Jenkins 1.609 is technically Java 6

src/main/java/jenkins/scm/api/trait/SCMSourceContext.java
```diff
@@ -0,0 +1,260 @@
...
+        return Collections.unmodifiableList(filters);
+    }
+
+    /**
+     * Returns the (possibly empty) list of {@link SCMSourceRequest} independent pre-filters.
+     *
+     * @return the (possibly empty) list of {@link SCMSourceRequest} independent pre-filters.
+     */
+    @NonNull
+    public final List<SCMHeadPrefilter> prefilters() {
```
stephenc: `prefilter` is a word, therefore camel-case would be `nonSense`

screen shot 2017-06-14 at 09 36 42

src/main/java/jenkins/scm/impl/trait/WildcardSCMHeadFilterTrait.java
stephenc: There should be a `GlobSCMHeadFilterTrait` at some point in the future, the acceptance criteria only requires the Wildcard for SCMHead and Regex for SCMSource and suggests to have one other implementation of each in order to prove that the API can have multiple implementations... so I completed the symmetry

src/main/resources/jenkins/scm/api/form/traits.jelly
stephenc: It would be nice, but there is a non-trivial diff:

screen shot 2017-06-14 at 09 46 22

In my opinion they are different

src/test/java/jenkins/scm/impl/SingleSCMSourceTest.java
stephenc: Yes it works, but I have switched to explicit to future-proof for when JENKINS-21017 is fixed

src/main/java/jenkins/scm/api/trait/SCMSourceBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Returns the {@link SCMSourceTrait} instances to provide to the {@link SCMSource} (assuming the {@link SCMSource}
+     * is {@link SCMSourceTrait} aware.
+     * @return the {@link SCMSourceTrait} instances to provide to the {@link SCMSource} (assuming the
+     * {@link SCMSource} is {@link SCMSourceTrait} aware.
+     */
+    @NonNull
+    public final List<SCMSourceTrait> traits() {
+        return Collections.unmodifiableList(traits);
```
stephenc: https://github.com/jenkinsci/scm-api-plugin/pull/36/commits/e4003a8365e613cba88f6e3cd0e3d379847bf691 should address your concerns

src/main/java/jenkins/scm/api/trait/SCMNavigatorRequest.java
```diff
@@ -0,0 +1,351 @@
...
+     * Constructor.
+     *
+     * @param source  the source.
+     * @param context  the context.
+     * @param observer the observer.
+     */
+    protected SCMNavigatorRequest(@NonNull SCMNavigator source, @NonNull SCMNavigatorContext<?, ?> context,
+                                  @NonNull SCMSourceObserver observer) {
+        this.navigator = source;
+        this.filters = new ArrayList<SCMSourceFilter>(context.filters());
```
stephenc: because a context can be reused to service multiple requests and the requests may be being processes in parallel (not currently for the 1 implementation, but rather bake it into the API up front)

src/main/java/jenkins/scm/api/trait/SCMHeadAuthority.java
```diff
@@ -0,0 +1,212 @@
...
+     * @return {@code true} if the supplied head is trusted.
+     * @throws IOException          if there was an I/O error trying to establish the trust status.
+     * @throws InterruptedException if interrupted while trying to establing the trust status.
+     */
+    @SuppressWarnings("unchecked")
+    public final boolean isTrusted(@NonNull SCMSourceRequest request, @NonNull SCMHead head)
+            throws IOException, InterruptedException {
+        return isApplicableTo(request)
+                && isApplicableTo(head)
+                && checkTrusted((S) request, (H) head);
```
stephenc: Ahh but they won't!

`isApplicableTo(request)` is `getDescriptor().isApplicableTo(request)` which ultimately ensures that the request is at least an instance of the parameterized type in the implementation's constructor.

same goes for `isApplicableTo(head)` so the casts will always work.

However that does raise the point that implementers should ensure they call the base method when overriding, hence 0b6a184e37bfc552b4636a13d94639e1756bc53f

src/main/java/jenkins/scm/api/trait/SCMSourceRequest.java
```diff
@@ -0,0 +1,553 @@
...
+     * Constructor.
+     *
+     * @param source   the source.
+     * @param context  the context.
+     * @param listener the (optional) {@link TaskListener}.
+     */
+    protected SCMSourceRequest(@NonNull SCMSource source, @NonNull SCMSourceContext<?, ?> context,
+                               @CheckForNull TaskListener listener) {
+        this.source = source;
+        this.filters = Collections.unmodifiableList(new ArrayList<SCMHeadFilter>(context.filters()));
```
stephenc: Yes, safety first

stephenc:
@rsandell are we good to go now?

PR #37, opened 2017-06-02
[FIXED JENKINS-44648] SCMRevisionAction should record corresponding source

stephenc:
See [JENKINS-44648](https://issues.jenkins-ci.org/browse/JENKINS-44648)

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMRevisionAction.java
+++ b/src/main/java/jenkins/scm/api/SCMRevisionAction.java
@@ -25,16 +25,26 @@ package jenkins.scm.api;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.RestrictedSince;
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/scm/api/SCMRevisionAction.java
```diff
@@ -44,11 +54,36 @@
     /**
      * Constructor.
      *
-     * @param rev the {@link SCMRevision}.
+     * @param revision the {@link SCMRevision}.
+     * @deprecated use {@link #SCMRevisionAction(SCMSource, SCMRevision)}
+     */
+    @Restricted(DoNotUse.class)
```
jglick: `@Restricted` on top of `@Deprecated` is just superfluous. Deprecation itself suffices for the intended purpose. Restrictions should be limited to newly introduced members which for technical reasons must be `public`.

stephenc: I want to stop people using this method

PR #38, opened 2017-06-14
[JENKINS-43507] Update implementation guide

stephenc:
@reviewbybees

### DIFF
--- a/docs/implementation.adoc
+++ b/docs/implementation.adoc
@@ -551,6 +551,12 @@ public class MySCMSource extends SCMSource {
      * non-mandatory fields should be non-final
      */
 
+    /**
+     * Using traits is not required but it does make your implementation easier for others to extend.
+     */
+    @NonNull
+    private List<SCMSourceTrait> traits = new ArrayList<>();
+
     @DataBoundConstructor
     public MockSCMSource(String id, /*mandatory configuration*/) {
         super(id); /* see note on ids*/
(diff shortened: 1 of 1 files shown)


stephenc:
Whitespace diff: https://github.com/jenkinsci/scm-api-plugin/pull/38/files?w=1

docs/implementation.adoc
```diff
@@ -562,6 +568,16 @@ public class MySCMSource extends SCMSource {
 
     // Getters for all the configuration fields
 
+    @NonNull
+    public List<SCMSourceTrait> getTraits() {
+        return new ArrayList<>(traits);
```
rsandell: `Collections.unmodifiableList(traits)` is what I've seen you use.

docs/implementation.adoc
```diff
@@ -711,8 +727,97 @@ public class MySCMSource extends SCMSource {
...
+    }
+
+    // provide builder-style getters and setters for fields
+
+    @NonNull
+    @Override
+    public MySCM build() {
+        MySCM result = new MySCM(this);
+        result.setHead(head(), revision());
+        // apply the decorations from the fields
```
rsandell: shouldn't you iterate the traits here and ask them to apply stuff to the scm?

stephenc: Nope, that is handled by the withRequest() call

docs/implementation.adoc
stephenc: ACK

PR #40, opened 2017-06-14
[FIXED JENKINS-44884] Add @Symbol annotations

stephenc:
See [JENKINS-44884](https://issues.jenkins-ci.org/browse/JENKINS-44884)

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/scm/impl/trait/WildcardSCMHeadFilterTrait.java
+++ b/src/main/java/jenkins/scm/impl/trait/WildcardSCMHeadFilterTrait.java
@@ -130,7 +130,7 @@ public class WildcardSCMHeadFilterTrait extends SCMSourceTrait {
     /**
      * Our descriptor.
      */
-    @Symbol("headsWildcardFilter")
+    @Symbol("headWildcardFilter")
     @Extension
     @Selection
     public static class DescriptorImpl extends SCMSourceTraitDescriptor {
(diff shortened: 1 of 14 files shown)


stephenc:
Downstream of https://github.com/jenkinsci/structs-plugin/pull/21

src/main/java/jenkins/scm/impl/trait/WildcardSCMHeadFilterTrait.java
```diff
@@ -129,6 +130,7 @@ private String getPattern(String branches) {
     /**
      * Our descriptor.
      */
+    @Symbol("headsWildcardFilter")
```
jglick: `headWildcardFilter` by analogy with `headRegexFilter` and following the general English rule (as I understand it) that this usage of “heads” is a kind of unmarked genitive

src/test/java/jenkins/scm/impl/SymbolAnnotationsTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Test
+    public void given__mockScm__when__uninstantiating__then__noRaw$class() throws Exception {
+        MockSCMController c = MockSCMController.create();
+        try {
+            c.createRepository("test");
+            MockSCM instance = new MockSCM(c, "test", new MockSCMHead("master"), new MockSCMRevision(
+                    new MockSCMHead("master"), c.getRevision("test", "master")));
+            assertThat(DescribableModel.uninstantiate2_(instance).toString(), allOf(
+                    startsWith("@"),
```
jglick: This use `TreeMap`s internally, so you can just assert an exact form for the `toString` and it should be stable.

jglick: Now that you have added the `is` match, the `allOf` matches seem redundant. (Same for other tests.)

stephenc:
> Now that you have added the `is` match, the `allOf` matches seem redundant. (Same for other tests.)

Actually I see the `is` match as redundant and the `allOf` is the actual criteria. If the `is` match breaks then we can understand what the parameters are in fixing the expected string value

PR #41, opened 2017-06-14
[FIXED JENKINS-44891] Migrate SCMSource's id field to a @DataBoundSetter

stephenc:
See [JENKINS-44891](https://issues.jenkins-ci.org/browse/JENKINS-44891)

@reviewbybees

Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/40

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -53,6 +53,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.TransientActionFactory;
 import net.jcip.annotations.GuardedBy;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * A {@link SCMSource} is responsible for fetching {@link SCMHead} and corresponding {@link SCMRevision} instances from
@@ -131,14 +133,65 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      *
      * @param id the id or {@code null}.
      */
+    @Deprecated
     protected SCMSource(@CheckForNull String id) {
         this.id = id == null ? UUID.randomUUID().toString() : id;
     }
 
     /**
-     * The ID of this source. The ID is not related to anything at all.
+     * Constructor.
+     */
+    protected SCMSource() {
+    }
+
+    /**
+     * Sets the ID that is used to ensure this {@link SCMSource} can be retrieved from its {@link SCMSourceOwner}
+     * provided that this {@link SCMSource} does not already {@link #hasId()}.
+     * <p>
+     * Note this is a <strong>one-shot</strong> setter. If {@link #getId()} is called first, then its value will stick,
+     * otherwise the first call to {@link #setId(String)} will stick.
+     *
+     * @param id the ID.
+     * @see #hasId()
+     * @see #getId()
+     * @since 2.2.0
+     */
+    @DataBoundSetter
+    public final synchronized void setId(@CheckForNull String id) {
+        if (this.id == null) {
+            this.id = id;
+        }
+    }
+
+    /**
+     * Variant of {@link #setId(String)} that can be useful for method chaining.
+     * @param id the ID
+     * @return {@code this} for method chaining
+     */
+    public final SCMSource withId(@CheckForNull String id) {
+        setId(id);
+        return this;
+    }
+
+    /**
+     * Returns {@code true} if and only if this {@link SCMSource} has been assigned an ID. Once an ID has been assigned
+     * it should be preserved.
+     *
+     * @return {@code true} if and only if this {@link SCMSource} has been assigned an ID.
+     * @see #setId(String)
+     * @see #getId()
+     * @since 2.2.0
+     */
+    public final synchronized boolean hasId() {
+        return this.id != null;
+    }
+    /**
+     * The ID of this source. The ID is not related to anything at all. If this {@link SCMSource} does not have an ID
+     * then one will be generated.
      *
      * @return the ID of this source.
+     * @see #setId(String)
+     * @see #hasId()
      */
     @NonNull
     public final synchronized String getId() {
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -151,7 +151,7 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      * Note this is a <strong>one-shot</strong> setter. If {@link #getId()} is called first, then its value will stick,
      * otherwise the first call to {@link #setId(String)} will stick.
      *
-     * @param id the ID.
+     * @param id the ID, this is an opaque token expected to be unique within any one {@link SCMSourceOwner}.
      * @see #hasId()
      * @see #getId()
      * @since 2.2.0
@@ -160,6 +160,8 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
     public final synchronized void setId(@CheckForNull String id) {
         if (this.id == null) {
             this.id = id;
+        } else if (!this.id.equals(id)) {
+            throw new IllegalStateException("The ID cannot be changed after it has been set.");
         }
     }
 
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -85,6 +85,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>structs</artifactId>
+      <version>1.8-20170614.081136-1</version> <!-- TODO remove SNAPSHOT before release -->
+      <!-- require at least 1.8 as existing implementations that add support for traits will need JENKINS-44864 -->
```
jglick: Well…this dep could be on 1.7, and the downstream PRs which actually use that idiom could request the newer version.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -131,14 +133,67 @@ public boolean equals(Object obj) {
...
+    @Deprecated
     protected SCMSource(@CheckForNull String id) {
         this.id = id == null ? UUID.randomUUID().toString() : id;
     }
 
     /**
-     * The ID of this source. The ID is not related to anything at all.
+     * Constructor.
+     */
+    public SCMSource() {
```
jglick: `protected`

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -131,14 +133,67 @@ public boolean equals(Object obj) {
...
+     * @see #hasId()
+     * @see #getId()
+     * @since 2.2.0
+     */
+    @DataBoundSetter
+    public final synchronized void setId(@CheckForNull String id) {
+        if (this.id == null) {
+            this.id = id;
+        } else {
+            System.out.println("Go away");
```
jglick: Uh, `IllegalStateException`?

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -131,14 +133,67 @@ public boolean equals(Object obj) {
...
+     * @param id the ID.
+     * @see #hasId()
+     * @see #getId()
+     * @since 2.2.0
+     */
+    @DataBoundSetter
+    public final synchronized void setId(@CheckForNull String id) {
+        if (this.id == null) {
+            this.id = id;
+        } else {
```
jglick:
```java
} else if (!id.equals(this.id)) {
```

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -131,14 +133,67 @@ public boolean equals(Object obj) {
...
+            System.out.println("Go away");
+        }
+    }
+
+    /**
+     * Variant of {@link #setId(String)} that can be useful for method chaining.
```
jglick: Do Stapler & `structs` not support returning values from `@DataBoundSetter`? Should be fixed if not.

src/main/java/jenkins/scm/impl/SingleSCMSource.java
```diff
@@ -79,14 +78,30 @@
...
+
+    /**
+     * Legacy constructor.
+     *
+     * @param id   source id.
+     * @param name the name of the branch.
+     * @param scm  the configuration.
+     * @deprecated use {@link #SingleSCMSource(String, SCM)} and {@link #setId(String)}.
+     */
+    @Restricted(NoExternalUse.class)
```
jglick: Please stop abusing `@Restricted`. A simple `@Deprecated` suffices.

src/test/java/jenkins/scm/impl/mock/MockSCMSource.java
```diff
@@ -72,28 +75,63 @@
     private final List<SCMSourceTrait> traits;
     private transient MockSCMController controller;
 
-    @DataBoundConstructor
+    @Restricted(NoExternalUse.class)
```
jglick: Completely ignored in test sources.

We make no guarantees about either binary or source compatibility of test utilities, so either leave it simply `@Deprecated`, if it can continue to work as before (which I think it can), or delete it if it cannot.

src/test/java/jenkins/scm/impl/mock/MockSCMSourceBuilder.java
```diff
@@ -52,8 +52,10 @@ public MockSCMSourceBuilder(String id, String controllerId, String repository) {
     @NonNull
     @Override
     public MockSCMSource build() {
-        return controller == null
-                ? new MockSCMSource(id, controllerId, repository, traits())
-                : new MockSCMSource(id, controller, repository, traits());
+        MockSCMSource source = controller == null
+                ? new MockSCMSource(controllerId, repository, traits())
+                : new MockSCMSource(controller, repository, traits());
+        source.setId(id);
```
jglick: Perhaps make `id` optional for `MockSCMSourceBuilder` too.

src/main/java/jenkins/scm/impl/SingleSCMSource.java
stephenc: I want to force people to upgrade the source when then upgrade the plugin

src/main/java/jenkins/scm/api/SCMSource.java
stephenc: Stapler does not IIRC

src/main/java/jenkins/scm/api/SCMSource.java
stephenc: test code

src/main/java/jenkins/scm/impl/SingleSCMSource.java
jglick: Why? If the old call still works and in the same way, this is just harassing plugin developers needlessly.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -131,14 +133,65 @@ public boolean equals(Object obj) {
...
+    }
+
+    /**
+     * Sets the ID that is used to ensure this {@link SCMSource} can be retrieved from its {@link SCMSourceOwner}
+     * provided that this {@link SCMSource} does not already {@link #hasId()}.
+     * <p>
+     * Note this is a <strong>one-shot</strong> setter. If {@link #getId()} is called first, then its value will stick,
+     * otherwise the first call to {@link #setId(String)} will stick.
+     *
+     * @param id the ID.
```
jglick: This would be a good place to document that it is an opaque token, expected to be unique within one owner, etc.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -131,14 +133,65 @@ public boolean equals(Object obj) {
...
+     * @see #hasId()
+     * @see #getId()
+     * @since 2.2.0
+     */
+    @DataBoundSetter
+    public final synchronized void setId(@CheckForNull String id) {
+        if (this.id == null) {
+            this.id = id;
+        }
```
jglick: Suggest

```
} else if (id != null && !id.equals(this.id)) {
    throw new IllegalStateException("attempt to clobber id");
}
```

src/test/java/jenkins/scm/impl/mock/MockSCMSource.java
```diff
@@ -73,27 +76,27 @@
     private transient MockSCMController controller;
 
     @DataBoundConstructor
-    public MockSCMSource(@CheckForNull String id, String controllerId, String repository, List<SCMSourceTrait> traits) {
-        super(id);
+    public MockSCMSource(String controllerId, String repository, List<SCMSourceTrait> traits) {
+        super();
```
jglick: can be omitted

src/test/java/jenkins/scm/impl/mock/MockSCMSourceBuilder.java
```diff
@@ -28,32 +28,43 @@
...
     @NonNull
     @Override
     public MockSCMSource build() {
-        return controller == null
-                ? new MockSCMSource(id, controllerId, repository, traits())
-                : new MockSCMSource(id, controller, repository, traits());
+        MockSCMSource source = controller == null
+                ? new MockSCMSource(controllerId, repository, traits())
+                : new MockSCMSource(controller, repository, traits());
+        if (id != null) {
```
jglick: harmless to call even if null, so could use `withId` IIUC

PR #42, opened 2017-07-06
[FIXED JENKINS-45331] Abstract classes should be abstract

stephenc:
See [JENKINS-45311](https://issues.jenkins-ci.org/browse/JENKINS-45331)

@reviewbybees


michaelneale:
:bee: once CI gives the go ahead.

PR #49, opened 2018-01-10
[FIXED JENKINS-48571] Ensure we always save after setting SCMSource id.

abayer:
[JENKINS-48571](https://issues.jenkins-ci.org/browse/JENKINS-48571)

I don't love using workflow-multibranch for the test, but that was the
easiest way I could find to reproduce the error. I'm also not sure if
the way I'm saving is actually the best approach.

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,12 +107,6 @@
       <version>1.6</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-multibranch</artifactId>
-      <version>2.16</version>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -65,8 +65,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>1.642.3</jenkins.version>
+    <java.level>7</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
 
@@ -105,6 +105,12 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>junit</artifactId>
       <version>1.6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-multibranch</artifactId>
+      <version>2.16</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -65,8 +65,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.642.3</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>1.609.3</jenkins.version>
+    <java.level>6</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
 
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -75,6 +75,8 @@ import org.kohsuke.stapler.DataBoundSetter;
 public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
         implements ExtensionPoint {
 
+    private static final Logger LOGGER = Logger.getLogger(SCMSource.class.getName());
+
     /**
      * Replaceable pronoun of that points to a {@link SCMSource}. Defaults to {@code null} depending on the context.
      * @since 2.0
@@ -197,10 +199,17 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      */
     @NonNull
     public final synchronized String getId() {
-        if (id == null) {
-            id = UUID.randomUUID().toString();
+        if (this.id == null) {
+            this.id = UUID.randomUUID().toString();
+            if (owner != null) {
+                try {
+                    owner.save();
+                } catch (IOException e) {
+                    LOGGER.log(Level.WARNING, "Could not save owner of SCMSource: " + e, e);
+                }
+            }
         }
-        return id;
+        return this.id;
     }
 
     /**
--- a/src/test/java/jenkins/scm/impl/PersistAcrossRestartsTest.java
+++ b/src/test/java/jenkins/scm/impl/PersistAcrossRestartsTest.java
@@ -3,28 +3,35 @@ package jenkins.scm.impl;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.model.FreeStyleProject;
-import hudson.model.InvisibleAction;
+import hudson.model.ItemGroup;
+import hudson.model.Job;
+import hudson.model.Run;
 import hudson.model.TaskListener;
+import hudson.model.TopLevelItem;
+import hudson.model.TopLevelItemDescriptor;
 import hudson.scm.NullSCM;
 import hudson.scm.SCM;
-import jenkins.branch.BranchSource;
+import jenkins.model.Jenkins;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMHeadEvent;
 import jenkins.scm.api.SCMHeadObserver;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMSource;
 import jenkins.scm.api.SCMSourceCriteria;
-import org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject;
+import jenkins.scm.api.SCMSourceOwner;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.RestartableJenkinsRule;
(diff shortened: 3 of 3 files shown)


pom.xml
```diff
@@ -65,8 +65,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
```
svanoort: Lol.

abayer: Yeaaaaaah. I just bumped it to match `workflow-multibranch`, though it obviously coulda gone higher.

pom.xml
```diff
@@ -107,6 +107,12 @@
       <version>1.6</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-multibranch</artifactId>
```
stephenc: Nope nope nope. We are not adding this. It has the potential to create circular dependency when using local snapshots. Just dual-purpose a top level item as scmsourceowner for tests

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -197,10 +199,17 @@ public final synchronized boolean hasId() {
      */
     @NonNull
     public final synchronized String getId() {
-        if (id == null) {
-            id = UUID.randomUUID().toString();
+        if (this.id == null) {
+            this.id = UUID.randomUUID().toString();
+            if (owner != null) {
```
stephenc: Ahhh... I wonder if we shouldn’t just assign the id during set owner instead

abayer: *shrug*

michaelneale:
nice catch! this was annoying

pom.xml
stephenc: We should be able to revert this now. I want to bump, but it will be preceded with an announcement and minor version bump.

src/main/java/jenkins/scm/api/SCMSource.java
stephenc: That was a request to try not an idle comment ;-)

src/test/java/jenkins/scm/impl/PersistAcrossRestartsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    public void uuidIdPersists() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                BogusSCMSource source = new BogusSCMSource();
+                SCMItem p = story.j.jenkins.createProject(SCMItem.class, "test-with-source");
+                p.setSource(source);
+                // TODO: Find a better way to persist this for testing across restart.
+                FreeStyleProject p2 = story.j.createFreeStyleProject("test-without-source");
+                p2.setDescription(source.getId());
```
stephenc: Side-effect invalidates test imho. I think we need to grab the ID by reflection or package scope in order to prove the issue is actually fixed

src/test/java/jenkins/scm/impl/PersistAcrossRestartsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @Test
+    public void uuidIdPersists() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                BogusSCMSource source = new BogusSCMSource();
+                SCMItem p = story.j.jenkins.createProject(SCMItem.class, "test-with-source");
+                p.setSource(source);
+                // TODO: Find a better way to persist this for testing across restart.
+                FreeStyleProject p2 = story.j.createFreeStyleProject("test-without-source");
```
stephenc: Just stash it in an AtomicReference in the Test class. The test class instance is durable for the duration of the test

src/test/java/jenkins/scm/impl/PersistAcrossRestartsTest.java
abayer: Huh? We need to actually call `getId()` after the item is saved. This fails without the fix, passes with it.

src/main/java/jenkins/scm/api/SCMSource.java
abayer: Well, my *shrug* was at least in part that I don't know if that's the right approach - it's entirely possible for someone to call `setId(String)` after `setOwner(..)` is called but before `getId()` is called.

stephenc: Once setOwner is called with non-null then the ID must be assigned if not already imho.

Now the question then becomes will the ID be assigned if saved *before* setOwner... and that is the rub...

May need to add a protected writeReplace that does the assignment... but then the test needs a pair of subclasses... one overriding and one not... or perhaps an XStream converter

abayer: I legitimately believe that the current implementation is the best (or at least least worst) yet discussed. It doesn't change any existing expectations (as autogenerating the ID on `setOwner(...)` if `id` is `null` would), it works fine with calling `setId(...)` or `getId()` before `setOwner(...)` is called, etc. Least complicated option is best option.

src/test/java/jenkins/scm/impl/PersistAcrossRestartsTest.java
```diff
@@ -0,0 +1,179 @@
...
+                SCMItem p = (SCMItem) story.j.jenkins.getItem("test-with-source");
+                assertNotNull(p);
+                SCMSource source = p.getSCMSource(uuid.get());
+                assertNotNull(source);
+                assertFalse(source instanceof NullSCMSource);
+            }
+        });
+    }
+
+    public static class BogusSCMSource extends SCMSource {
```
jglick: `MockSCMSource` does not suffice?

src/main/java/jenkins/scm/api/SCMSource.java
jglick: Why not leave this here, but also add a `transient` flag indicating that the ID had been generated; and then in `setOwner`, if that flag is set, save the owner then also? I think that covers all the bases without getting too complicated. OTOH it seems likely that a `setOwner` call would have already been accompanied by a save.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -197,10 +199,17 @@ public final synchronized boolean hasId() {
...
+                try {
+                    owner.save();
+                } catch (IOException e) {
+                    LOGGER.log(Level.WARNING, "Could not save owner of SCMSource: " + e, e);
+                }
+            }
         }
-        return id;
+        return this.id;
```
jglick: No need for `this.` qualifiers unless the field is masked by a local variable.

stephenc:
To remove doubt. Until I say otherwise: DO NOT MERGE.

stephenc:
I’ll have some time in airports next week, so I can give this problem a good look.

stephenc:
This is not the way to fix this issue. I have the bones of the correct way implemented but I have not had time to finish it off yet.

Closing this PR to clear up any doubt in case people get confused.

maxime-michel:
@stephenc can you maybe open a PR with what you have so people can contribute or at least track the progress? Thanks.

alexsuter:
Is there any other workaround? After every reboot of my jenkins master I have to rescan all my builds.

maxime-michel:
I personally have set an interval of a day for reindexing pipeline jobs' branches. On Jenkins' JIRA some people are saying this triggers new builds, it doesn't for us. (I'm not sure why because I've seen that behavior as well.) In any case, the interval isn't respected very precisely but so far we don't get red builds when we merge a PR after a couple of days of inactivity, which is what we needed.

michaelneale:
hey @maxime-michel @alexsuter - I believe this was closed as it is still a bit fiddly one to automatically fix. Basically jobs created via jobDSL (or anything else, like blue ocean) need to assign an ID now. Are you able to to try the following to confirm it helps (I am assuming you are using jobDSL, is that right?)

A work around is a small change as described by @samrocketman: 
https://github.com/samrocketman/jervis/commit/e4cd6324ff22c3593d7e6feab88dff79e516e14b


```
Stephen Connolly your suggestion appears to have fixed the issue.
Required
After making the change from Stephen Connolly's suggestion I had to click "Scan Repository Now".  The problem continued to be broken until I scanned the repository for branches.
```

see comment: https://issues.jenkins-ci.org/browse/JENKINS-46290?focusedCommentId=329162&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-329162


Sorry about all this, this looks like it does strange things when not set, and we want to make sure any fix doesn't create "build storms"

maxime-michel:
Thanks for the details, however, I'm actually not using Job DSL.

michaelneale:
@maxime-michel how did you create the job?

alexsuter:
@michaelneale Thanks. But I always use blue ocean and declarative build pipelines.

michaelneale:
@alexsuter oh in that case https://github.com/jenkinsci/blueocean-plugin/pull/1662 will help (for new ones). For existing you should be able to open the config screen, and save it (no change) and from then on things should be normal (perhaps there is more going on, but that was a known problem talked about in a few channels).

maxime-michel:
Thanks again for the help @michaelneale. I haven't seen the error in a couple of days and I did save all jobs (to enable the scan interval as I was saying previously).

stephenc:
@maxime-michel you may need to trigger a full scan after saving to flush out any references to the unsaved id that was only in memory

michaelneale:
@maxime-michel thank @stephenc tingling spider sense to know something was up with the id on restart...

PR #50, opened 2018-01-26
Discussion for AccuRev Implementation

casz:
Created PR to start a discussion about implementation for AccuRev.

My thoughts so far are to create SCM Source with the following.
Should I implement new HeadCategory to easier different between Workspaces and Staging streams?
For instance, for `StagingStreamDiscoveryTrait` you might want to consider workspaces directly below the Staging Stream depending on the build strategy on the trait. Yet ignore all other workspaces in the stream tree.
You might have something committed to the workspace (AccuRev terms kept or defunct) which would be the reason to build workspaces below a staging stream.

And for `WorkspaceDiscoveryTrait` you will definitely ignore workspaces connected to staging streams yet build rest.

Still, have a lot of work to do with implementing webhook sent and receive on the AccuRev server and share that 😢

### DIFF
--- a/docs/implementation.adoc
+++ b/docs/implementation.adoc
@@ -357,6 +357,18 @@ public class AccurevSCMSource extends SCMSource {
                     new NonLocalizable("Streams")
                     // Better: Messages._AccurevSCMSource_StreamHeadCategory()
                 ),
+                new UncategorizedSCMSourceCategory(
+                    new NonLocalizable("Gated Streams")
+                    // Better: Messages._AccurevSCMSource_GatedStreamHeadCategory()
+                ),
+                new ChangeRequestSCMHeadCategory(
+                    new NonLocalizable("Staging streams")
+                    // Better: Messages._AccurevSCMSource_StreamHeadCategory()
+                )
+                new ChangeRequestSCMHeadCategory(
+                    new NonLocalizable("Workspace")
+                    // Better: Messages._AccurevSCMSource_StreamHeadCategory()
+                ),
                 new TagSCMHeadCategory(
                     new NonLocalizable("Snapshots")
                     // Better: Messages._AccurevSCMSource_SnapshotHeadCategory()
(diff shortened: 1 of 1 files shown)


docs/implementation.adoc
```diff
@@ -357,6 +357,18 @@ public class AccurevSCMSource extends SCMSource {
                     new NonLocalizable("Streams")
                     // Better: Messages._AccurevSCMSource_StreamHeadCategory()
                 ),
+                new UncategorizedSCMSourceCategory(
```
stephenc: you can't have more than one `UncategorizedSCMSourceCategory`

docs/implementation.adoc
```diff
@@ -357,6 +357,18 @@ public class AccurevSCMSource extends SCMSource {
...
+                    new NonLocalizable("Gated Streams")
+                    // Better: Messages._AccurevSCMSource_GatedStreamHeadCategory()
+                ),
+                new ChangeRequestSCMHeadCategory(
```
stephenc: These will need to have their own category classes... you can host them in the accurev plugin as they only make sense there.

If you want you should (i.e. I have not checked) be able to *extend* `ChangeRequestSCMHeadCategory` if they share the semantics of a change request

docs/implementation.adoc
```diff
@@ -357,6 +357,18 @@ public class AccurevSCMSource extends SCMSource {
...
+                    new NonLocalizable("Staging Streams")
+                    // Better: Messages._AccurevSCMSource_StagingStreamHeadCategory()
+                )
+                new ChangeRequestSCMHeadCategory(
```
stephenc: I think workspaces deserve their own category... though I am unsure if you really want them.

OTOH it would be cool to enable a dev to get CI reports of their local changes

casz: That's my exact thought too, having it as a trait but not by default I think would be awesome 👍

docs/implementation.adoc
casz: A gated stream is exactly like a stream however what happens underneath is different. A gated stream spawns staging stream for each changing stream that is underneath and wants to promote changes. Could be a workspace or a regular stream or even another gated stream is what's

PR #51, opened 2018-03-13
Mention space separator in exclusions by name help file

vgaidarji:
### What has been done
It's not very obvious how to specify multiple name patterns. Adding this information to help file should help.

💡 ~Should we update the pattern example as well?~ - YES

### DIFF
--- a/src/main/resources/jenkins/scm/impl/trait/WildcardSCMSourceFilterTrait/help-excludes.html
+++ b/src/main/resources/jenkins/scm/impl/trait/WildcardSCMSourceFilterTrait/help-excludes.html
@@ -22,6 +22,6 @@
  ~ THE SOFTWARE.
  -->
 <div>
-    Project name patterns to ignore even if matched by the includes list.
+    Project name patterns (separated by space) to ignore even if matched by the includes list.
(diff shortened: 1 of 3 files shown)


vgaidarji:
@casz done 👍

src/main/resources/jenkins/scm/impl/trait/WildcardSCMSourceFilterTrait/help-excludes.html
```diff
@@ -22,6 +22,6 @@
  ~ THE SOFTWARE.
  -->
 <div>
-    Project name patterns to ignore even if matched by the includes list.
-    For example: <code>release</code>
+    Project name patterns (separated by space) to ignore even if matched by the includes list.
+    For example: <code>release alpha-* beta-*</code>
```
jglick: Copy-pasted example, was misleading. Presumably these should look like names of projects, rather than names of branches.

vgaidarji: Yeah, good point. I see this pattern used in includes help file as well, fixed.

PR #54, opened 2018-05-11
Improve WildcardSCMHeadFilterTrait/help-excludes.html message

ikesyo:
Let's make it similar to help-includes.html.


ikesyo:
Can someone take a look? Maybe @jglick or @stephenc

PR #55, opened 2018-08-21
Add default traits and setter

abayer:
I'm honestly not sure why these weren't already here.

cc @rsandell @dwnusbaum

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -170,7 +170,7 @@ public abstract class SCMNavigator extends AbstractDescribableImpl<SCMNavigator>
      * Gets the traits for this navigator.
      * @return traits the list of traits, empty by default.
      */
-    @CheckForNull
+    @NonNull
     public List<SCMTrait<? extends SCMTrait<?>>> getTraits() {
         return Collections.emptyList();
     }
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -37,10 +37,6 @@ import hudson.model.Items;
 import hudson.model.TaskListener;
 import hudson.util.AlternativeUiTextProvider;
 import hudson.util.LogTaskListener;
-import jenkins.model.TransientActionFactory;
-import jenkins.scm.api.trait.SCMTrait;
-import org.kohsuke.stapler.DataBoundSetter;
-
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Iterator;
@@ -49,6 +45,9 @@ import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import jenkins.model.TransientActionFactory;
+import jenkins.scm.api.trait.SCMTrait;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * An API for discovering new and navigating already discovered {@link SCMSource}s within an organization.
--- a/src/main/java/jenkins/scm/api/SCMNavigator.java
+++ b/src/main/java/jenkins/scm/api/SCMNavigator.java
@@ -37,6 +37,10 @@ import hudson.model.Items;
 import hudson.model.TaskListener;
 import hudson.util.AlternativeUiTextProvider;
 import hudson.util.LogTaskListener;
+import jenkins.model.TransientActionFactory;
+import jenkins.scm.api.trait.SCMTrait;
+import org.kohsuke.stapler.DataBoundSetter;
+
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Iterator;
@@ -45,8 +49,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import jenkins.model.TransientActionFactory;
-import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * An API for discovering new and navigating already discovered {@link SCMSource}s within an organization.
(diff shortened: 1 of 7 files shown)


abayer:
@stephenc @rsandell - could one of you review and hopefully merge/release this? Thanks.

src/test/java/jenkins/scm/impl/mock/MockSCMSource.java
```diff
@@ -362,6 +369,8 @@ public ListBoxModel doFillRepositoryItems(@QueryParameter String controllerId) t
             return SCMSourceTrait._for(this, MockSCMSourceContext.class, MockSCMBuilder.class);
         }
 
+        @Override
+        @Nonnull
```
stephenc: Wrong Nonnull annotation

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -37,6 +37,10 @@
 import hudson.model.TaskListener;
 import hudson.util.AlternativeUiTextProvider;
 import hudson.util.LogTaskListener;
+import jenkins.model.TransientActionFactory;
+import jenkins.scm.api.trait.SCMTrait;
```
stephenc: A-Z no grouping

src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java
```diff
@@ -27,13 +27,16 @@
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.model.Descriptor;
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.Set;
```
stephenc: A-Z no grouping

src/test/java/jenkins/scm/impl/mock/MockSCMNavigator.java
```diff
@@ -52,6 +46,13 @@
 import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
```
stephenc: A-Z no grouping

src/test/java/jenkins/scm/impl/mock/MockSCMNavigator.java
```diff
@@ -52,6 +46,13 @@
...
 
+import javax.annotation.Nonnull;
```
stephenc: Where is the published JSR defining a spec for a `javax` annotation of this name?

abayer:
Imports and annotation updated.

abayer:
@stephenc ping for re-review?

stephenc:
Did you address my query on symbols generically?

abayer:
No, because that seems to be a separate thing from just adding these methods.

stephenc:
I think it would be a mistake to significantly push traits further after discovering the @symbol name conflict issue without finding a way for two sources to have different implications of the same trait symbol name and not have the symbol support blow up and force use of the FQCN despite the case that the implications are unambiguously assignable

stephenc:
If you are critically blocked on this, then go ahead... but my belief is the symbol problem is intrinsically linked (as once I hit it I stopped pushing beyond git based SCMs (which at least share a common parent class)

abayer:
I'm definitely not disagreeing re: there being an issue with symbol resolution - I'm just saying that adding no-op traits getter/setter methods doesn't actually create any problems. I need this specifically to be able to do instantiation of a multibranch project with an arbitrary `SCMSource`, and since you can't actually create a functional multibranch project using `GitSCMSource` et al without manually setting the default traits, my alternative to this PR is to do reflection crap. So yeah, we do need to move forward on this - either it's gonna get merged/released ASAP or I need to change my code elsewhere to do reflection ugliness.

stephenc:
Bitbucket and GitHub (And Gitea And in future GitLab) all want their own “discover fork PRs” trait with different options and backing implementation.

We’d like each one to be able to call the trait the same name, eg `discoverForkPRs` but as soon as we do that you cannot instantiate the trait using symbol... (try configuring the `resolveScm` step in the Pipeline Syntax Web UI)

So instead, we end up with `discoverGitHubForkPRs`, `discoverBitbucketForkPRs`, etc... which is not good.

I think a “symbol helper” extension of some sort that can filter the extension list contextually might be the solution... that way scm-api could register one and use the bag of trait descriptors to limit the options and we’d only have the GitHub option for GitHub so it could “share” the same trait name for the equivalent functionality

stephenc:
Speaking of bag of trait descriptors (I’m on phone and cannot check) you do add the trait descriptors bag to the abstract SCMSourceDescriptor base class?

abayer:
Yes, I did. Let me know when you've created the JIRA.

stephenc:
You may be waiting some time if you’re waiting on me to create the jira

abayer:
Fine by me - if you're waiting for me to do it, you're gonna wait even longer, since you're the one who actually knows what you want. =)

stephenc:
sad panda
https://issues.jenkins-ci.org/browse/JENKINS-53825

abayer:
@stephenc So are we ok to merge this, then?

stephenc:
If @rsandell agrees

src/main/java/jenkins/scm/api/SCMNavigator.java
```diff
@@ -158,6 +159,22 @@ protected final void resetId() {
...
+     * @param traits the list of traits
+     */
+    public void setTraits(@CheckForNull List<SCMTrait<? extends SCMTrait<?>>> traits) {
+    }
+
+    /**
+     * Gets the traits for this navigator.
+     * @return traits the list of traits, empty by default.
+     */
+    @CheckForNull
```
rsandell: Eeh, in the docs it says `@NonNull`?

abayer: whoops! Fixing.

PR #56, opened 2018-10-01
[JENKINS-53752] Add ability for SCMHeadAuthority to exclude from discovery

abayer:
[JENKINS-53752](https://issues.jenkins-ci.org/browse/JENKINS-53752)

Will very shortly be upstream of https://github.com/jenkinsci/github-branch-source-plugin/pull/188

Oh, and I incrementalized while I was here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.2.9</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>1.609.3</jenkins.version>
+    <jenkins.version>2.7.3</jenkins.version>
     <java.level>7</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -29,12 +29,12 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.36</version>
+    <version>3.21</version>
     <relativePath />
   </parent>
 
   <artifactId>scm-api</artifactId>
-  <version>2.2.9-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
   <packaging>hpi</packaging>
 
   <name>SCM API Plugin</name>
@@ -61,12 +61,14 @@
     <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-    <tag>HEAD</tag>
+    <tag>${scmTag}</tag>
   </scm>
 
   <properties>
+    <revision>2.2.9</revision>
+    <changelist>-SNAPSHOT</changelist>
     <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
+    <java.level>7</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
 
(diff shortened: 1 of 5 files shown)


src/main/java/jenkins/scm/api/trait/SCMHeadAuthority.java
```diff
@@ -147,6 +147,15 @@ protected boolean checkTrusted(@NonNull S request, @NonNull R revision) throws I
         return checkTrusted(request, (H) revision.getHead());
     }
 
+    /**
+     * Checks if the authority is configured to allow discovery of untrusted {@link SCMHead}s.
+     *
+     * @return true by default for compatibility. Override in an implementation to change this.
```
stephenc: How will this apply if there are multiple authorities as indicated in the class javadoc?

Feels like the wrong place to expose this. What other places did you consider

pom.xml
```diff
@@ -29,12 +29,12 @@
...
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.36</version>
+    <version>3.21</version>
     <relativePath />
   </parent>
 
   <artifactId>scm-api</artifactId>
-  <version>2.2.9-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
```
stephenc: Can we do incrementals in a separate PR

src/main/java/jenkins/scm/api/trait/SCMHeadAuthority.java
abayer: If any are false, that would win out, same as with `SCMHeadFilter` etc.

Over on the original PR, you asked me to put this on the individual authorities, so...that's what I'm doing, but bumping it up to `SCMHeadAuthority` so that we can reuse the same logic/API elsewhere too.

pom.xml
abayer: I needed this to publish an incremental that I can consume in the downstream.

stephenc: Yep. Just put it in a separate PR and we’ll merge that first

src/main/java/jenkins/scm/api/trait/SCMHeadAuthority.java
stephenc: I wonder if we'd be better serviced with something like `getImplicitFilters()` that returns the filters that apply.

On the other hand the way you are implementing this, the result only applies when this authority is defining the head as untrusted.

Basically my spider-sense is tingling and saying this is a hack into the API that we will regret later.

Can you explore some other approaches (even just on-paper) and comment back with a summary and we'll see what fits best

abayer: Other approaches like what? I don't know this code base near well enough to come up with anything but what at least on the surface appear to be hacks. And yes, the result only applies when it's an untrusted head, but that's by design - that's literally the functionality that's been requested.

stephenc: So from what I can see, we actually may need to migrate the authority API to a new primary method.

we currently have checkTrusted which returns a boolean.

We actually need at least some form of tri-state. This could be `Optional<Boolean>` or it could be an enum.

* One value says "I can state for certain that this head is trusted"
* One value says "I can state for certain that this head is not trusted"
* One value says "I no nothing about this head"

We could perhaps go further with a 4 valued enum, e.g.

* Definitively trusted (stop checking all the rest)
* Assume trusted (continue checking but allowed to report trusted unless another says Untrusted)
* Unknown
* Untrusted

Which of these to pick requires thinking through.

The outcome should be that we know which authoritie(s) are responsible for the head... and it could legitimately be a plural.

Tell me again why this needs to be in SCM-API and why this cannot be a property of GitHub branch source just deciding during a scan to ignore untrusted if a flag has been set on the context? (would also save API calls)

abayer: I assumed this would be behavior that would be desirable on more `SCMSource`s than just `GitHubSCMSource`, so it made sense to do it here so that other `SCMSource` implementations just have to override `isDiscoverUntrusted()` in their `SCMHeadAuthority`s.

I have basically no clue what you're talking about in re: more values for "is this trusted", so I'm guessing I should probably step away from this.

jglick:
Downstream was closed so I suppose this is abandoned.

PR #58, opened 2018-10-03
[JENKINS-52964] Add SCMFileSystem.supports(SCMSourceDescriptor)

abayer:
[JENKINS-52964](https://issues.jenkins.io/browse/JENKINS-52964)

It's a pain that we can't tell whether a given `SCMSource`
implementation can support an `SCMFileSystem` without having an actual
instance of the `SCMSource` - that makes it impossible, for example,
to filter a list of possible `SCMSource`s in a configuration UI to
just those which support `SCMFileSystem`.

This is a proposed way to address that - add
`SCMFileSystem.Builder#supports(SCMSourceDescriptor)`, and
`SCMFileSystem#supports(SCMSourceDescriptor)`. The `Builder` method
will return false unless overridden, to preserve compatibility, and
it's up to individual `Builder` implementations to decide whether they
want to say "yes, the corresponding `SCMSource` for this
`SCMSourceDescriptor` is supported" without an actual instance to
check against for more details.

(downstream PRs consuming this incoming shortly)

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMFileSystem.java
+++ b/src/main/java/jenkins/scm/api/SCMFileSystem.java
@@ -364,6 +364,31 @@ public abstract class SCMFileSystem implements Closeable {
     }
 
     /**
+     * Given a {@link SCMSourceDescriptor} this method will check if there is at least one {@link SCMFileSystem} provider
+     * capable of being instantiated from the descriptor's {@link SCMSource}. Returning {@code true} does not mean that
+     * {@link #of(SCMSource, SCMHead, SCMRevision)} will be able to instantiate a {@link SCMFileSystem} for any specific
+     * {@link SCMHead} or {@link SCMRevision}, rather returning {@code false} indicates that there is absolutely no point
+     * in calling {@link #of(SCMSource, SCMHead, SCMRevision)} as it will always return {@code null}.
+     *
+     * @param descriptor the {@link SCMSourceDescriptor}.
+     * @return {@code true} if {@link #of(SCMSource, SCMHead)} / {@link #of(SCMSource, SCMHead, SCMRevision)} could
+     * return a {@link SCMFileSystem} implementation, {@code false} if {@link #of(SCMSource, SCMHead)} /
+     * {@link #of(SCMSource, SCMHead, SCMRevision)} will always return {@code null} for the supplied {@link SCMSource}.
+     * @since 2.0
+     */
+    public static boolean supports(@NonNull SCMSourceDescriptor descriptor) {
+        descriptor.getClass(); // throw NPE if null
+        if (descriptor.clazz == null) {
+            throw new NullPointerException();
+        }
+        for (Builder b : ExtensionList.lookup(Builder.class)) {
+            if (b.supports(descriptor)) {
+                return true;
+            }
+        }
+        return false;
+    }
+    /**
      * Extension point that allows different plugins to implement {@link SCMFileSystem} classes for the same {@link SCM}
      * or {@link SCMSource} and let Jenkins pick the most capable for any specific {@link SCM} implementation.
      */
@@ -387,6 +412,18 @@ public abstract class SCMFileSystem implements Closeable {
          * {@code null}.
          */
         public abstract boolean supports(SCMSource source);
+
+        /**
+         * Checks if this {@link Builder} supports the supplied {@link SCMSourceDescriptor}.
+         *
+         * @param descriptor the {@link SCMSourceDescriptor}
+         * @return {@code true} if and only if the supplied {@link SCMSourceDescriptor}'s {@link SCMSource} class is
+         * supported by this {@link Builder}, {@code false} if {@link #build(SCMSource, SCMHead, SCMRevision)} will
+         * <strong>always</strong> return {@code null}, and {@code false} by default for compatibility.
+         */
+        public boolean supports(SCMSourceDescriptor descriptor) {
+            return false;
+        }
 
         /**
          * Given a {@link SCM} this should try to build a corresponding {@link SCMFileSystem} instance that
--- a/src/main/java/jenkins/scm/api/SCMFileSystem.java
+++ b/src/main/java/jenkins/scm/api/SCMFileSystem.java
@@ -34,6 +34,7 @@ import hudson.model.Item;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.scm.SCM;
+import hudson.scm.SCMDescriptor;
 import hudson.scm.SCMRevisionState;
 import java.io.Closeable;
 import java.io.File;
@@ -365,6 +366,31 @@ public abstract class SCMFileSystem implements Closeable {
     }
 
     /**
+     * Given a {@link SCMDescriptor} this method will check if there is at least one {@link SCMFileSystem} provider
+     * capable of being instantiated from the descriptor's {@link SCMSource}. Returning {@code true} does not mean that
+     * {@link #of(Item, SCM, SCMRevision)} will be able to instantiate a {@link SCMFileSystem} for any specific
+     * {@link Item} or {@link SCMRevision}, rather returning {@code false} indicates that there is absolutely no point
+     * in calling {@link #of(Item, SCM, SCMRevision)} as it will always return {@code null}.
+     *
+     * @param descriptor the {@link SCMDescriptor}.
+     * @return {@code true} if {@link #of(Item, SCM, SCMRevision)} could return a {@link SCMFileSystem} implementation,
+     * {@code false} if {@link #of(Item, SCM, SCMRevision)} will always return {@code null} for the supplied {@link SCM}.
+     * @since 2.3.0
+     */
+    public static boolean supports(@NonNull SCMDescriptor descriptor) {
+        descriptor.getClass(); // throw NPE if null
+        if (descriptor.clazz == null) {
+            throw new NullPointerException();
+        }
+        for (Builder b : ExtensionList.lookup(Builder.class)) {
+            if (b.supports(descriptor)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Given a {@link SCMSourceDescriptor} this method will check if there is at least one {@link SCMFileSystem} provider
      * capable of being instantiated from the descriptor's {@link SCMSource}. Returning {@code true} does not mean that
      * {@link #of(SCMSource, SCMHead, SCMRevision)} will be able to instantiate a {@link SCMFileSystem} for any specific
@@ -375,7 +401,7 @@ public abstract class SCMFileSystem implements Closeable {
      * @return {@code true} if {@link #of(SCMSource, SCMHead)} / {@link #of(SCMSource, SCMHead, SCMRevision)} could
      * return a {@link SCMFileSystem} implementation, {@code false} if {@link #of(SCMSource, SCMHead)} /
      * {@link #of(SCMSource, SCMHead, SCMRevision)} will always return {@code null} for the supplied {@link SCMSource}.
-     * @since 2.0
+     * @since 2.3.0
      */
     public static boolean supports(@NonNull SCMSourceDescriptor descriptor) {
         descriptor.getClass(); // throw NPE if null
@@ -389,6 +415,7 @@ public abstract class SCMFileSystem implements Closeable {
         }
         return false;
     }
+
     /**
      * Extension point that allows different plugins to implement {@link SCMFileSystem} classes for the same {@link SCM}
      * or {@link SCMSource} and let Jenkins pick the most capable for any specific {@link SCM} implementation.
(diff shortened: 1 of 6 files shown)


src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -363,6 +363,31 @@ public static boolean supports(@NonNull SCMSource source) {
...
+     * {@link SCMHead} or {@link SCMRevision}, rather returning {@code false} indicates that there is absolutely no point
+     * in calling {@link #of(SCMSource, SCMHead, SCMRevision)} as it will always return {@code null}.
+     *
+     * @param descriptor the {@link SCMSourceDescriptor}.
+     * @return {@code true} if {@link #of(SCMSource, SCMHead)} / {@link #of(SCMSource, SCMHead, SCMRevision)} could
+     * return a {@link SCMFileSystem} implementation, {@code false} if {@link #of(SCMSource, SCMHead)} /
+     * {@link #of(SCMSource, SCMHead, SCMRevision)} will always return {@code null} for the supplied {@link SCMSource}.
+     * @since 2.0
+     */
+    public static boolean supports(@NonNull SCMSourceDescriptor descriptor) {
```
stephenc: If we need this for the `SCMSource`'s `Descriptor` then we need it for the `SCM`'s `Descriptor` too

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -388,6 +413,18 @@ public static boolean supports(@NonNull SCMSource source) {
...
+        /**
+         * Checks if this {@link Builder} supports the supplied {@link SCMSourceDescriptor}.
+         *
+         * @param descriptor the {@link SCMSourceDescriptor}
+         * @return {@code true} if and only if the supplied {@link SCMSourceDescriptor}'s {@link SCMSource} class is
+         * supported by this {@link Builder}, {@code false} if {@link #build(SCMSource, SCMHead, SCMRevision)} will
+         * <strong>always</strong> return {@code null}, and {@code false} by default for compatibility.
+         */
+        public boolean supports(SCMSourceDescriptor descriptor) {
+            return false;
```
stephenc: I think this might not be the correct value. Ideally what we would do is make this method `final` and make the `SCMSource` version non-abstract. Then we introduce `protected abstract` methods that the `public final` methods call (guarded by a reflective check of non-implemented)

This would allow us to try and infer if the answer *should* be `true` or `false` for a non-implemented method based on things like:
* Is it in the same package to the `SCMSource`
* Is the `SCMSource` an outer-class of this `Builder`

I think if either of those are the case and we are dealing with a legacy `Builder` implementation then we should return `true` and not `false`, otherwise we will have false negatives and you basically have to instantiate the `SCMSource` anyway

abayer: That's reasonable - I did probably get the default value wrong. I'll think on the rest.

stephenc: To be clear, we only check heuristics if the `protected abstract boolean checkSupports(SCMSourceDescriptor);` has not been implemented. Which would be the case only for plugins compiled against the older versions of scm-api. As soon as you bump the scm-api dependency, you'd be forced to implement it and thus can return the correct `true`/`false` value and the heuristics can safely be ignored

stephenc:
NOTE: this change will require a MINOR version bump after merge

src/main/java/jenkins/scm/api/SCMFileSystem.java
stephenc: Symmetry.

Plus I can think of quite a few SCM's that would know the answer as being `true` out of the box:
* Subversion
* Accurev (if they ever implement scm-api)

It's really only Git/Mercurial that need to know more about how the repo is hosted before a good answer can be provided... and for that matter same answer applies to `GitSCMSource`

src/main/java/jenkins/scm/api/SCMFileSystem.java
abayer: The heuristics will also need to handle inheritance - i.e., a specific `AbstractGitSCMSource` implementation may not have its own `SCMFileSystem.Builder` implementation, but it can still use `GitSCMFileSystem`.

abayer: Wait, maybe not - since we're checking all `SCMFileSystem.Builder` implementations, we just need the `GitSCMFileSystem.Builder#supportsDescriptor` implementation to be implemented in that scenario.

abayer: (all told, this just really makes me wish we had the `SCMSource` as a type parameter on `SCMFileSystem.Builder` or something like that. Ah well.)

src/main/java/jenkins/scm/api/SCMFileSystem.java
abayer: Alrightie!

abayer:
@stephenc - added `supports(SCMDescriptor)` and related, switched to the `final`/`protected abstract`/heuristics combo. Lemme know what you think?

src/test/java/jenkins/scm/impl/SingleSCMSourceTest.java
```diff
@@ -220,6 +224,14 @@ public void getSCMDescriptors() throws Exception {
                 (Matcher) hasItem(instanceOf(MockSCM.DescriptorImpl.class)));
     }
 
+    @Issue("JENKINS-52964")
+    @Test
+    public void filesystem_supports_false_by_default_for_descriptor() {
```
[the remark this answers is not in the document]
abayer: There's testing of the positive case down in https://github.com/jenkinsci/git-plugin/pull/623 if that helps - otherwise, lemme know what other scenarios you'd like to see tests for.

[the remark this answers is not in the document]
stephenc: Scm-api should have positive and negative test cases. Relying on a downstream repo will not prevent broken releases.

Also in case I forgot, MINOR version bump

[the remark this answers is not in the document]
abayer: Gotcha, will do.

And I already did bump the version to 2.3.0 after you brought it up before. =)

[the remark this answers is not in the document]
stephenc: > And I already did bump the version to 2.3.0 after you brought it up before. =)

ack. it's hard to review on a phone!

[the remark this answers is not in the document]
abayer: Fun, this is going to require a couple mock `SCMFileSystem` implementations - I'll try to get that done tomorrow.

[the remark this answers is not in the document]
stephenc: Just extend the mock with some other implementations, registered with `@TestExtension` and then remove the mock we have's builder from the extension list when you want to test

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -388,6 +441,67 @@ public static boolean supports(@NonNull SCMSource source) {
...
+         * a child package of the {@link SCM}.
+         * @since 2.3.0
+         */
+        public final boolean supports(SCMDescriptor<?> descriptor) {
+            if (Util.isOverridden(SCMFileSystem.Builder.class, getClass(), "supportsDescriptor", SCMDescriptor.class)) {
+                return supportsDescriptor(descriptor);
+            } else {
+                return descriptor.clazz.isAssignableFrom((getClass().getEnclosingClass())) ||
+                        getClass().getPackage().equals(descriptor.clazz.getPackage()) ||
+                        getClass().getPackage().getName().startsWith(descriptor.clazz.getPackage().getName());
```
jglick: ulch

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -388,6 +441,67 @@ public static boolean supports(@NonNull SCMSource source) {
...
+         *
+         * @param descriptor the {@link SCMDescriptor}
+         * @return the return value of {@link #supportsDescriptor(SCMDescriptor)} if implemented, otherwise, it
+         * will return true if one of the following is true: the {@link SCM} for the descriptor is the enclosing
+         * class of this builder class, the builder and {@link SCM} are in the same package, or the builder is in
+         * a child package of the {@link SCM}.
+         * @since 2.3.0
+         */
+        public final boolean supports(SCMDescriptor<?> descriptor) {
+            if (Util.isOverridden(SCMFileSystem.Builder.class, getClass(), "supportsDescriptor", SCMDescriptor.class)) {
```
jglick: Weird. Just make the method non`abstract` and provide the `else` clause as the default implementation. Why is this overcomplicated?

abayer: @stephenc wanted it this way. =)

src/main/java/jenkins/scm/api/SCMFileSystem.java
abayer: I am soooo open to a better implementation or heuristic.

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -388,6 +441,67 @@ public static boolean supports(@NonNull SCMSource source) {
...
+
+        /**
+         * Checks if this {@link Builder} supports the supplied {@link SCMDescriptor}.
+         *
+         * @param descriptor the {@link SCMDescriptor}
+         * @return {@code true} if and only if the supplied {@link SCMSourceDescriptor}'s {@link SCMDescriptor} class is
+         * supported by this {@link Builder}.
+         * @since 2.3.0
+         */
+        protected abstract boolean supportsDescriptor(SCMDescriptor descriptor);
```
rsandell: Won't this break any existing `Builder` implementations that hasn't yet adapted when upgrading?!

stephenc: If the logic is correct and the access is guarded, it should only cause compiler errors on updating the dependency... which is what we'd want, as if we could guess the answer we wouldn't need the method.

Of course the question for code review is has this been done correctly

abayer: That does raise @jglick's question - what's the reasoning for this approach versus a single non-abstract method with the default heuristic that implementations can override?

stephenc: When I was doing my PhD my supervisor was notorious for redlining drafts of papers. People would routinely go through 50 or 70 revisions of their papers.

For my first paper, after 15 odd revisions I started to notice that every even numbered revision was basically the same and the odd numbered revisions were basically the same...

The redlines with corrections were oscillating back and forth... so revision 16 I didn’t take his corrections verbatim but instead used it as a hint to rewrite the sentence from scratch. Version 18 was the one we sent for publication and my subsequent papers were only 4-5 revisions.

You seem to be blindly following suggestions I make on my phone rather than stepping back and trying to figure out why i’m Prodding at these things (much like I was just taking my supervisor’s redlines verbatim)

I hope this story helps you

abayer: ...so I guess I took “Ideally what we would do is make this method final and make the SCMSource version non-abstract. Then we introduce protected abstract methods that the public final methods call (guarded by a reflective check of non-implemented)” too literally? And that doesn’t answer my question as to why you wanted this approach in the first place. So, no, the story is not helpful. =)

stephenc: So one of the lessons I learned in SCM-API is that there is great benefit in separation of the API from the SPI.

In SCM API the SPI (i.e. what implementers are supposed to implement) should be protected methods (to prevent consumers from calling them) whereas the API can be public methods on the SCM API API.

Given that we want to change/enhance behaviour for this method, that says that this is no longer a simple API, so we'd want to migrate more fully to the SPI vs API separation.

SPI evolution is different from API evolution. the SPI can break compile compatibility as long as runtime behaviour falls through. So in general that means ensuring that there is only one abstract method for anything rather than 5 overloads that people have to puzzle which to implement. If we need to add parameters, we move the `abstract` from one `protected` method to another. Then the `public final` method does the disambiguation and falls back through the potential implementations until it finds one that works or returns the "sensible" response.

abayer:
Well, if there's no good reason not to do it in a simpler way, I'm doing it in the simpler way, so I've switched to just the one method with a default implementation.

src/main/java/jenkins/scm/api/SCMFileSystem.java
abayer: so...does that mean I should rollback to the final/abstract combo? My instinct is to go with a simple single overrideable method - but this is your turf, not mine, so I am not in a position to dictate the path forward, you are.

stephenc:
I want the method the user invokes to be `public final` and the method that the implementor provides is either:

* `protected abstract`; or
* in a separate class that users are not supposed to instantiate or access themselves

abayer:
Ok, then I'll revert my last change.

abayer:
@rsandell - there's nothing in `docs` regarding the `Builder#supports` methods in the first place, so I'm not sure if anything needs to be added.

PR #61, opened 2019-01-07
Using own executor service

ingwarsw:
In our company we have Jenkins with ~ 1000 repos in GitHub.
Most of them have a lot of tags (> 1k is standard, moved from old SCMs, but we cannot remove them so far)
Because of that processing events can take a loot of time, and Timer pool is really often all being used for event, which turns our jenkins into vegetable.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.4.2</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.107.1</jenkins.version>
     <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -65,9 +65,9 @@
   </scm>
 
   <properties>
-    <revision>2.4.2</revision>
+    <revision>2.5.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.107.1</jenkins.version>
+    <jenkins.version>2.164.3</jenkins.version>
     <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
--- a/src/main/java/jenkins/scm/api/SCMEvent.java
+++ b/src/main/java/jenkins/scm/api/SCMEvent.java
@@ -28,10 +28,14 @@ package jenkins.scm.api;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionList;
+import hudson.init.Terminator;
 import hudson.model.Cause;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.security.ACL;
+import hudson.util.ClassLoaderSanityThreadFactory;
+import hudson.util.DaemonThreadFactory;
+import hudson.util.NamingThreadFactory;
 import java.util.Date;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
@@ -40,10 +44,6 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.servlet.http.HttpServletRequest;
-
-import hudson.util.ClassLoaderSanityThreadFactory;
-import hudson.util.DaemonThreadFactory;
-import hudson.util.NamingThreadFactory;
 import jenkins.security.ImpersonatingScheduledExecutorService;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
@@ -215,6 +215,18 @@ public abstract class SCMEvent<P> {
             executorService = new ImpersonatingScheduledExecutorService(new ScheduledThreadPoolExecutor(10, new NamingThreadFactory(new ClassLoaderSanityThreadFactory(new DaemonThreadFactory()), "SCMEvent")), ACL.SYSTEM);
         }
         return executorService;
+    }
+
+
+    /**
+     * Shutdown the timer and throw it away.
+     */
+    @Terminator
+    public synchronized void closeExecutorService() {
+        if (executorService != null) {
+            executorService.shutdownNow();
+            executorService = null;
+        }
     }
 
     /**
@@ -467,6 +479,7 @@ public abstract class SCMEvent<P> {
         }
 
         protected abstract void log(SCMEventListener l, Throwable e);
+
         protected abstract void fire(SCMEventListener l, E event);
 
         @Override
--- a/src/main/java/jenkins/scm/api/SCMEvent.java
+++ b/src/main/java/jenkins/scm/api/SCMEvent.java
@@ -34,20 +34,21 @@ import hudson.model.TaskListener;
 import hudson.security.ACL;
 import java.util.Date;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.servlet.http.HttpServletRequest;
-import jenkins.util.Timer;
+
+import hudson.util.DaemonThreadFactory;
+import hudson.util.NamingThreadFactory;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
 import org.apache.commons.lang.StringUtils;
-import org.jvnet.localizer.Localizable;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.Stapler;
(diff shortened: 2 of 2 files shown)


ingwarsw:
Any thoughts about this?
It would be nice to include it someday, do I need to add some extra data?

bitwiseman:
@ingwarsw 
I'm not vary familiar with this area, but what you did here lines up with the "future-proofing" comment.  😄 
It looks reasonable to me, but we should get another review before merging.

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -200,10 +206,13 @@ protected SCMEvent(SCMEvent<P> copy) {
...
     @NonNull
-    protected static ScheduledExecutorService executorService() {
-        // Future-proofing, if we find out that events are drowning Timer then we may need to move them to their
-        // own dedicated ScheduledExecutorService thread pool
-        return Timer.get();
+    protected static synchronized ScheduledExecutorService executorService() {
+        if (executorService == null) {
+            // corePoolSize is set to 10, but will only be created if needed.
+            // ScheduledThreadPoolExecutor "acts as a fixed-sized pool using corePoolSize threads"
+            executorService =  new ScheduledThreadPoolExecutor(10, new NamingThreadFactory(new DaemonThreadFactory(), "SCMEvent"));
```
jglick: Note that `Timer` is also wrapped in utilities to

* impersonate `ACL.SYSTEM`
* handle errors, if we are not using `SafeTimerTask`
* maintain the `Thread.contextClassLoader`

Whether any of those issues are important for this usage, I have no idea offhand. But perhaps `Timer` should get a utility factory

```java
public static ScheduledExecutorService newService(String name, int corePoolSize) {
    return new ImpersonatingScheduledExecutorService(new ErrorLoggingScheduledThreadPoolExecutor(corePoolSize, new NamingThreadFactory(new ClassLoaderSanityThreadFactory(new DaemonThreadFactory()), name)), ACL.SYSTEM);
}
```

for easy reuse.

ingwarsw: This one does everything except logging errors if we are not using `SafeTimerTask`

I can use exactly the same as in timer... but I choose to use that one for some purpose..
It was so long ago I dont even remember why.. 
But I remember it was not accident..

ingwarsw: And `Timer` is in core so I cannot change it from here..

jglick: > This one does everything except logging errors

It does not do the impersonation or context class loader maintenance.

> `Timer` is in core so I cannot change it from here

Of course, but we can have a core PR, and via [JEP-305](https://jenkins.io/jep/305) it is even straightforward to have an `on-hold` PR in the plugin which verifies its usage.

ingwarsw: I see that there was change in Timer.. ok.. so it does more now..

So it would be great to move it somewhere in one place..

And maybe it would be easier and better to just allow changing size of timer pool via variable or something?
Here https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/util/Timer.java#L47

jglick: Again, to be clear, I am not _requesting_ that the executor service here add all of those protections, I am just noting for purposes of review that it does not.

> maybe it would be easier and better to just allow changing size of timer pool

Perhaps, but that is independent of this issue. The `Timer` pool is intended to be used for quick, nonblocking tasks. The implication of this PR is that some SCM event handlers were not quick. There is no associated JIRA issue or example thread dump so I cannot evaluate whether they were _intended_ to be quick and are not due to some bug, or whether it is just natural that they sometimes block for a while and thus should never have been using `Timer` to begin with.

ingwarsw: They are not quick.. they sometimes takes hours to complete.. (partially because of work that they need to do or of other bugs that I will try to talk about maybe on gitter, I need to ask how we would like to resolve them)

So pool size in timer is irrelevant in this context..

And about that added protections Im not sure also.. this local pool does one rather specified task, so maybe they are not needed.
But to be sure I would like to add them here as they are in Timer at current moment.

jglick: > they sometimes takes hours to complete

A thread dump would be the background we need to analyze why.

ingwarsw: I can get thread dump.. just need to wait for busy day..
But all waits here https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/RateLimitHandler.java#L39
We are just got rate limit badly..

ingwarsw: And while we dont have problems with waiting for GH.. 
We have problem that nothing else (that relies on `Timer`) on jenkins is working..

jglick: Need to thread dump to see what _caller_ is to blame.

ingwarsw: Webhook processing is going into this pool..
And there is issue with GH processing itself (sometimes rate limiting is not resetting)..
But not that easy to grasp..
And would be best to fix somewhere here..
https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GitHub.java#L316
Once for good :)

ingwarsw:
Guys I can update it to looks exactly like in `Timer` but I will have to update jenkins core version `2.60.3` -> `2.107.1` to have all required classes.

stephenc:
Go for the core version bump. Seems entirely reasonable to bump to that LTS at this time. Because of the core version bump, also bump the plugin’s minor version and reset patch to 0

ingwarsw:
@stephenc Done..

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -34,12 +34,17 @@
...
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.servlet.http.HttpServletRequest;
-import jenkins.util.Timer;
+
+import hudson.util.ClassLoaderSanityThreadFactory;
```
stephenc: all imports in a-z order, no blank lines please

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -198,10 +208,13 @@ protected SCMEvent(SCMEvent<P> copy) {
...
-        // own dedicated ScheduledExecutorService thread pool
-        return Timer.get();
+    protected static synchronized ScheduledExecutorService executorService() {
+        if (executorService == null) {
+            // corePoolSize is set to 10, but will only be created if needed.
+            // ScheduledThreadPoolExecutor "acts as a fixed-sized pool using corePoolSize threads"
+            executorService = new ImpersonatingScheduledExecutorService(new ScheduledThreadPoolExecutor(10, new NamingThreadFactory(new ClassLoaderSanityThreadFactory(new DaemonThreadFactory()), "SCMEvent")), ACL.SYSTEM);
+        }
+        return executorService;
     }
```
stephenc: we'll need a `@Terminator` annotated static method that shuts down this pool to play nicely with servlet containers that get reused, e.g. if anyone is in a downstream plugin and has lots of event unit tests, they may end up with *lots* of these thread pools. Better to expressly shut them down with the Jenkins lifecycle. 

See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/init/Terminator.java

ingwarsw:
@stephenc `Terminator in that form?

src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -198,10 +208,25 @@ protected SCMEvent(SCMEvent<P> copy) {
...
+        }
+        return executorService;
+    }
+
+
+    /**
+     * Shutdown the timer and throw it away.
+     */
+    @Terminator
+    public synchronized void closeExecutorService() {
```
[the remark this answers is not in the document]
ingwarsw: Actually I didnt think about it..

And I removed static because Im not sure if `@Terminator` can work on static method..
But if works we should add that static back to be on safe side..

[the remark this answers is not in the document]
ingwarsw: I see even @stephenc mentioned it should be static..
I will create fast fixing PR..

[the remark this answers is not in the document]
ingwarsw: https://github.com/jenkinsci/scm-api-plugin/pull/68

pom.xml
```diff
@@ -65,9 +65,9 @@
   </scm>
 
   <properties>
-    <revision>2.4.2</revision>
+    <revision>2.5.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.164.3</jenkins.version>
```
dwnusbaum: FWIW usually preferable to update to the oldest LTS line that has whatever core API you need. Especially in this case, because the plugin is an API intended to be used by other plugins, requiring the second-newest LTS release means that any plugin that wants to pick up the latest version of this plugin also requires the that release, which is a bit prohibitive.

jglick: In this case `2.107.3` was the minimum required baseline as far as I know.

PR #64, opened 2019-04-05
Don't fail SCMName detection on non-IDN host names

twz123:
Previously, when passing an URL with a host name that is not a valid IDN to SCMName::from, it failed with an IllegalArgumentException. This especially affects the case when the host name is an IP address. Make this method a bit more graceful by returning null instead.


MarkEWaite:
The CI failure is due to a JDK 11 javadoc problem on the master branch which will be resolved when #66 is merged.

twz123:
Thanks!

PR #65, opened 2019-04-24
[JENKINS-43802] Allow fetch/fetchRevisions to be given an Item context

jglick:
Needed for [JENKINS-43802](https://issues.jenkins-ci.org/browse/JENKINS-43802).

I thought of adding a builder field to `SCMSourceContext` / `SCMSourceRequest` but these seem heavily tied to branch scanning, whereas this is a different use case, for libraries. I also thought of adding something like

```java
/**
 * Allows this to be associated with a particular context, such as a job in which a build is running.
 * If defined, should be considered to supersede {@link #getOwner}
 * for purposes of {@link #retrieve(String, TaskListener)}, {@link #retrieveRevisions(TaskListener)}, etc.
 * @param context a context, such as a folder or job
 * @return {@code this} by default; may be overridden to create a <em>new</em> object with a copy of this configuration
 */
public @NonNull SCMSource withContext(@CheckForNull Item context) {
    return this;
}
```

which could even have a default implementation using either `DescribableModel` or `Cloneable` to make a copy of the original object, offering a `protected Item getContext()` to subclasses, but this seemed to just be perpetuating the antipattern of having critical parameters to a method be injected magically from some `owner` field, so I decided to keep it simple and just introduce new overloads for what I needed. Alternate design suggestions would certainly be considered.

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -609,11 +609,31 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      * @throws IOException if an error occurs while performing the operation.
      * @throws InterruptedException if any thread has interrupted the current thread.
      * @since 1.3
+     * @deprecated rather call {@link #fetch(String, TaskListener, Item)}
      */
+    @Deprecated
     @CheckForNull
     public final SCMRevision fetch(@NonNull String thingName, @CheckForNull TaskListener listener)
             throws IOException, InterruptedException {
         return retrieve(thingName, defaultListener(listener));
+    }
+
+    /**
+     * Looks up a specific thingName based on some SCM-specific set of permissible syntaxes.
+     * Delegates to {@link #retrieve(String, TaskListener)}.
+     * @param thingName might be a branch name, a tag name, a cryptographic hash, a change request number, etc.
+     * @param listener the task listener (optional)
+     * @param context an associated context to supersede {@link #getOwner}, such as a job in which this is running
+     * @return a valid {@link SCMRevision} corresponding to the argument, with a usable corresponding head, or
+     * {@code null} if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since 2.5
+     */
+    @CheckForNull
+    public final SCMRevision fetch(@NonNull String thingName, @CheckForNull TaskListener listener, @CheckForNull Item context)
+            throws IOException, InterruptedException {
+        return retrieve(thingName, defaultListener(listener), context);
     }
 
     /**
@@ -626,13 +646,36 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      * @throws IOException if an error occurs while performing the operation.
      * @throws InterruptedException if any thread has interrupted the current thread.
      * @since 1.3
+     * @deprecated rather override {@link #retrieve(String, TaskListener, Item)}
      */
+    @Deprecated
     @CheckForNull
     protected SCMRevision retrieve(@NonNull final String thingName, @NonNull TaskListener listener)
             throws IOException, InterruptedException {
+        if (Util.isOverridden(SCMSource.class, getClass(), "retrieve", String.class, TaskListener.class, Item.class)) {
+            return retrieve(thingName, listener, getOwner());
+        }
         SCMHeadObserver.Named baptist = SCMHeadObserver.named(thingName);
         _retrieve(null, baptist, null, listener);
         return baptist.result();
+    }
+
+    /**
+     * SPI: Looks up a specific revision based on some SCM-specific set of permissible syntaxes.
+     * The default implementation uses {@link #retrieve(SCMSourceCriteria, SCMHeadObserver, TaskListener)}
+     * and looks for {@link SCMHead#getName} matching the argument (so typically only supporting branch names).
+     * @param thingName might be a branch name, a tag name, a cryptographic hash, a revision number, etc.
+     * @param listener the task listener
+     * @param context an associated context to supersede {@link #getOwner}, such as a job in which this is running
+     * @return a valid revision object corresponding to the argument, with a usable corresponding head, or null if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since 2.5
+     */
+    @CheckForNull
+    protected SCMRevision retrieve(@NonNull final String thingName, @NonNull TaskListener listener, @CheckForNull Item context)
+            throws IOException, InterruptedException {
+        return retrieve(thingName, listener);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stephenc:
You are number 2 in the review queue (https://github.com/jenkinsci/credentials-plugin/pull/114 is number 1) it may be a few days before I can start processing the review queue

jglick:
> it may be a few days

Well the bug turns two years old today so I guess there is no urgency!

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -626,15 +646,38 @@ public final SCMRevision fetch(@NonNull String thingName, @CheckForNull TaskList
...
+     * @param context an associated context to supersede {@link #getOwner}, such as a job in which this is running
+     * @return a valid revision object corresponding to the argument, with a usable corresponding head, or null if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since 2.5
+     */
+    @CheckForNull
+    protected SCMRevision retrieve(@NonNull final String thingName, @NonNull TaskListener listener, @CheckForNull Item context)
+            throws IOException, InterruptedException {
+        return retrieve(thingName, listener);
```
stephenc: I feel like we could do better here, e.g.

* If we add a `ThreadLocal` that _if present_ is used as the return value of `getOwner` we could set it to the context item and remove it in a try-finally block around this SPI method and thus existing implementations would gain the functionality for free. 
* A more heavy weight approach might just be to switch the `owner` field - it's fully encapsulated already - for this kind of call (on the basis that implementations might be using a thread pool for some operations... and lazily looking up authentication in the thread pool... so a ThreadLocal wouldn't work anyway)

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -609,13 +609,33 @@ protected SCMRevision retrieve(@NonNull SCMHead head, @NonNull TaskListener list
...
+     * @param listener the task listener (optional)
+     * @param context an associated context to supersede {@link #getOwner}, such as a job in which this is running
+     * @return a valid {@link SCMRevision} corresponding to the argument, with a usable corresponding head, or
+     * {@code null} if malformed or not found
+     * @throws IOException if an error occurs while performing the operation.
+     * @throws InterruptedException if any thread has interrupted the current thread.
+     * @since 2.5
+     */
+    @CheckForNull
+    public final SCMRevision fetch(@NonNull String thingName, @CheckForNull TaskListener listener, @CheckForNull Item context)
```
stephenc: If this change makes sense for this API entry point then it makes sense for all.

That is:

```
    protected abstract void retrieve(@CheckForNull SCMSourceCriteria criteria,
                            @NonNull SCMHeadObserver observer,
                            @CheckForNull SCMHeadEvent<?> event,
                            @NonNull TaskListener listener)
            throws IOException, InterruptedException;
```

should probably be moved to a variant that includes the context item. That is the main implementation point. 

The idea is this: the API provides some well-known use cases, such as getting a specific revision of some specific thing. Implementations do not need to implement for all these use cases, instead they can start by implementing the main SPI points (i.e. the `abstract` ones) and they will get an implementation of the common use cases "for free"... of course the free implementation is not optimized, so implementations can improve things by overriding the various SPI optimization points.

What you are doing with this change is addressing your use case only, not the generic need. As such this change in its current format would:

1. Require all implementations to provide a context aware SPI override of a hidden method (it's non-abstract)
2. Not actually work against implementations that do not provide the hidden method.

bitwiseman: @stephenc 
What you're describing sound like a significantly different design and of larger scope as well.  

Perhaps that would be separate JIRA issue to be done later?

src/main/java/jenkins/scm/api/SCMSource.java
bitwiseman: @stephenc 
The `ThreadLocal` idea sounds exactly like the kind of "magical" thing that Jesse was trying to avoid.  What am I misunderstanding?

bitwiseman:
@stephenc
What you're describing sounds like a significantly larger scope of change.
 
Does the proposed change address JENKINS-43802?    If so, perhaps we could file a JIRA to do the larger scoped change at later date?

stephenc:
@bitwiseman in my experience, every time we "just add this one method to fix this one issue" rather than integrate the change holistically into the API, we have ended up badly burned... at least in terms of SCM-API.

So the changes requested are IMHO the minimum changes to get this change merged

jglick:
> What you are doing with this change is addressing your use case only, not the generic need.

Indeed. However I see no indication that others consumers of this plugin might ever need this `Item context`. The usual usage is for multibranch projects, which set an `owner` field. This use case is actually independent: library retrievers do not have an `SCMSourceOwner`.

> As such this change in its current format would:
> 
> 1. Require all implementations to provide a context aware SPI override

They are not _required_ to provide it. An old implementation may continue to override the non-context-aware overload, if it does not care about the context anyway.

> of a hidden method (it's non-abstract)

Indeed this aspect of compatible migrations is unpleasant. WDYT about an `@Abstract` annotation with corresponding processor that would make it a compile-time error to not override a given method?

> 2. Not actually work against implementations that do not provide the hidden method.

See above. If an implementation overrides the old SPI overload, any context passed by a caller is ignored. Conversely, if the caller calls the old API overload, an implementation will see a null context. That is the purpose of the `Util.isOverridden` tricks. There is nothing to be improved in that regard so far as I can tell. Anyway I have already patched all of the implementations _and_ the one caller to use the new overloads in downstream PRs, meaning anyone who simply updates their plugins will pick up the fix. So I would disregard the comments about `ThreadLocal`s and `legacyOwner`s and so on as being inapplicable.

If you continue to block the current API proposal on the grounds that in the future someone _might_ find a use case for passing a different sort of context to other `SCMSource` methods (those not used by library retrievers), then I am not sure what to say. We could deprecate the existing `SCMSource` methods used only by library retrievers and create a completely different `Describable` type for this purpose, with a matching library retriever implementation, which would be fine from an API perspective but would be problematic for user settings unless we found some trick to `readResolve` the old configurations. Should I prototype something along those lines, or would you prefer to discuss options live?

bitwiseman:
@jglick 
Then I'm in favor of the design as it is.

stephenc:
https://github.com/jglick/scm-api-plugin/pull/1

bitwiseman:
@stephenc 
That PR is rife with extraneous changes.  Could you revert the stuff that doesn't apply to this PR so we can see what you're really suggesting?

stephenc:
@bitwiseman I do not have any bandwidth to work on this further. Hiding whitespace changes will reduce the noise from keeping the changes files formatted correctly

jglick:
> Hiding whitespace changes will reduce the noise

Unfortunately GitHub diffs only support the equivalent `git diff -w`, which does not cover multiline reformattings like

```diff
-     * Fetches the latest heads and corresponding revisions. Implementers are free to cache intermediary results
-     * but the call must always check the validity of any intermediary caches.
+     * Fetches the latest heads and corresponding revisions. Implementers are free to cache intermediary results but the
+     * call must always check the validity of any intermediary caches.
```

jglick:
> so we can see what you're really suggesting

At a high level: injecting `@CheckForNull Item context` parameters into most major API/SPI methods in the plugin, for consistency and in case some use case might be found in the future; whereas my current patch injects them only into the methods which actually need it for purposes of library retrieval. My response was in https://github.com/jenkinsci/scm-api-plugin/pull/65#issuecomment-495770801.

stephenc:
From my point of view, there is importance in symmetry of the api otherwise it devolves into layer upon layer of hacks. Having had to spend the greater part of a year previously restoring a semblance of symmetry after permitting unmonitored changes to the API, I am enforcing that API changes are more designed rather than just quick hacks.

bitwiseman:
@jglick 
I've filed a PR to @stephenc 's PR that cleans up the changes he's requesting.
https://github.com/stephenc/scm-api-plugin/pull/1 

The resulting change to this PR looks like this: 
https://github.com/jglick/scm-api-plugin/compare/SCMSource.owner-JENKINS-43802...bitwiseman:jenkins-43802?expand=1

jglick:
> I've filed a PR to @stephenc's PR

OMG 3rd-level PRs, DVCS FTW

jglick:
> The resulting change to this PR looks like this

Seems better from a very brief glance.

stephenc:
I’d prefer if this was changed to a PR on @jglick’s PR so that we can retain the delta of other changes.

stephenc:
Three levels of PR is overly excessive. Better yet, just pull them all into one and create a new PR on the root repo directly

bitwiseman:
@stephenc 
No, thanks.  I'm offering a cleaned up version of your changes as a courtesy.  It is still your PR.

jglick:
We can do both. GitHub lets you submit the same commit in multiple branches as multiple PRs. You can look at things whichever way you want.

vivek:
>Three levels of PR is overly excessive. Better yet, just pull them all into one and create a new PR on the root repo directly

@stephenc all it needs for you is to merge @bitwiseman's changes in your PR. Its pretty much your PR with some clean up. These are not his changes, he did it as favor to speed things up. Let's move on this team!

bitwiseman:
After extensive discussion and review, we have decided to merge this change.

PR #68, opened 2019-06-20
Fixing synchronisation on termination of execution service

ingwarsw:


stephenc:
I don't have time to cut the release *and* update the wiki page. can you update the wiki page

PR #67, opened 2019-06-20
[JENKINS-43802] Allow fetch/fetchRevisions to be given an Item context

stephenc:
See [JENKINS-43802](https://issues.jenkins-ci.org/browse/JENKINS-43802).

This is #65 with the three-layers of nested PRs flattened

PR #70, opened 2019-06-24
[JENKINS-43802] Alternative implementation

stephenc:
Alternative approach to [JENKINS-43802](https://issues.jenkins-ci.org/browse/JENKINS-43802).

As I see it, there are multiple potential valid use cases for reusing `SCMSource` in a context where we cannot convert the owning `Item` into a `SCMSourceOwner`. As such we should provide a mechanism to adapt an `Item` into a `SCMSourceOwner`.

There are some implications to this:

* We need consumers of this new API to decide whether they want their sources discoverable by the event system or not. For example [JENKINS-41497](https://issues.jenkins-ci.org/browse/JENKINS-41497) indicates that a significant number of users probably do not want implicit triggering of all jobs using a shared library when the shared library is changed.

* We need to specifically support credentials lookup on the proxy, which means adding a dependency to the credentials plugin. The user documentation already guides implementers to use the credentials plugin, so that shouldn't be an issue.

### DIFF
--- a/docs/consumer.adoc
+++ b/docs/consumer.adoc
@@ -242,6 +242,21 @@ TIP: In plain english, when you trigger a build of a job for a `SCMRevision` you
 +
 For example, a Git `SCMSource` might return _Repository_, an Accurev `SCMSource` might return _Depot_, a CVS `SCMSource` might return _Module_ while a Subversion `SCMSource` could return _Repository_ or perhaps even something more generic like _Project_ depending on the way in which the Subversion server is being used footnote:[Do you create one `trunk|branches|tags` per repository or do you use a single big repository and create many `_project_/(trunk|branches|tags)` in that single repository?].
 
+[NOTE]
+====
+There are some rare but legitimate cases where a `SCMSource` needs to be owned by a `Item` that cannot extend `SCMSourceOwner`.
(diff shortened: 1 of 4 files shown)


jglick:
This kind of deep reflective magic seems like a really bad idea. #65, or even its proposed expansion as #67, is much more transparent and straightforward.

stephenc:
Has the advantages:

* none of the SCM implementations need updating
* enables the use cases that #67 enables without forcing a lot of API changes
* Proxy is not "deep" reflective magic, rather a standard API used quite widely. (I agree that it is a kind of magic... just disagree that it is deep magic) Could have been replaced with a simple `Item` delegation concrete class but that risks hindering core evolution.

stephenc:
The code could be a lot "simpler" if we avoid enabling the SCMEvent support, that's what adds complexity, but given that I suspect there are valid use cases (e.g. I know some on the Apache Maven project would actually want to trigger all jobs that use a shared library when the shared library is updated.. I disagree with them, and I suspect that once INFRA slaps them back after the first time... but still users may have a valid use case) that are contingent to adding the `SCMEvent` support... consequently I see the additional code as required

bitwiseman:
See #65.

PR #71, opened 2019-07-02
JENKINS-58313 Expose source commit

henryju:
Relates to https://groups.google.com/d/topic/jenkinsci-dev/-ojqklsaKbw/discussion

### DIFF
--- a/src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
+++ b/src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
@@ -1,0 +1,45 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package jenkins.scm.api.mixin;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * Additional attributes of a {@link ChangeRequestSCMHead} that should have been in the original mixin but we are not
+ * targeting Java 8 so we cannot add the default methods to the interface and must have an ugly {@code 3} class instead.
+ * @since 2.5.0
+ */
+// TODO once Java 8 is baseline move method to ChangeRequestSCMHead with default return value,
+// TODO deprecate this interface and add @Restricted(NoExternalUse.class) (retain empty interface for binary compat)
+public interface ChangeRequestSCMHead3 extends ChangeRequestSCMHead2 {
+
+    /**
+     * Returns the commit identifier (e.g. sha1) of the source branch.
+     *
+     * @return the commit identifier of the source branch of this change request or @null if such a concept doesn't exists in the backing source control system.
+     */
+    @CheckForNull
+    String getSourceCommitId();
+}
--- a/src/test/java/jenkins/scm/impl/mock/MockChangeRequestSCMHead.java
+++ b/src/test/java/jenkins/scm/impl/mock/MockChangeRequestSCMHead.java
@@ -26,12 +26,12 @@ package jenkins.scm.impl.mock;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Locale;
+import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMHeadOrigin;
 import jenkins.scm.api.mixin.ChangeRequestCheckoutStrategy;
-import jenkins.scm.api.SCMHead;
-import jenkins.scm.api.mixin.ChangeRequestSCMHead2;
+import jenkins.scm.api.mixin.ChangeRequestSCMHead3;
 
-public class MockChangeRequestSCMHead extends SCMHead implements ChangeRequestSCMHead2 {
+public class MockChangeRequestSCMHead extends SCMHead implements ChangeRequestSCMHead3 {
     private final String target;
     private final Integer number;
     private final SCMHeadOrigin origin;
@@ -39,7 +39,7 @@ public class MockChangeRequestSCMHead extends SCMHead implements ChangeRequestSC
 
     public MockChangeRequestSCMHead(SCMHeadOrigin origin, Integer number, String target,
                                     ChangeRequestCheckoutStrategy strategy, boolean singleStrategy) {
-        super("CR-"+number + (singleStrategy ? "" : "-"+strategy.name().toLowerCase(Locale.ENGLISH)));
+        super("CR-" + number + (singleStrategy ? "" : "-" + strategy.name().toLowerCase(Locale.ENGLISH)));
         this.number = number;
         this.target = target;
         this.origin = origin;
@@ -83,5 +83,10 @@ public class MockChangeRequestSCMHead extends SCMHead implements ChangeRequestSC
         // MockSCMController maintains a separate list of change-requests independent from the branches
         // so unlike GitHub or Bitbucket, the name of a change request in the origin is the name.
         return getName();
+    }
+
+    @Override
+    public String getSourceCommitId() {
+        return "sha" + number;
     }
 }


src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
```diff
@@ -0,0 +1,45 @@
...
+ * THE SOFTWARE.
+ */
+
+package jenkins.scm.api.mixin;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * Additional attributes of a {@link ChangeRequestSCMHead} that should have been in the original mixin but we are not
+ * targeting Java 8 so we cannot add the default methods to the interface and must have an ugly {@code 3} class instead.
```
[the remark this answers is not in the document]
jglick: Yes we should use `default` interface methods going forward.

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
```diff
@@ -0,0 +1,45 @@
...
+ * @since 2.5.0
+ */
+// TODO once Java 8 is baseline move method to ChangeRequestSCMHead with default return value,
+// TODO deprecate this interface and add @Restricted(NoExternalUse.class) (retain empty interface for binary compat)
+public interface ChangeRequestSCMHead3 extends ChangeRequestSCMHead2 {
+
+    /**
+     * Returns the commit identifier (e.g. sha1) of the source branch.
+     *
+     * @return the commit identifier of the source branch of this change request or @null if such a concept doesn't exists in the backing source control system.
```
jglick:
```suggestion
     * @return the commit identifier of the source branch of this change request or {@code null} if such a concept doesn't exists in the backing source control system.
```

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
```diff
@@ -0,0 +1,45 @@
...
+ * Additional attributes of a {@link ChangeRequestSCMHead} that should have been in the original mixin but we are not
+ * targeting Java 8 so we cannot add the default methods to the interface and must have an ugly {@code 3} class instead.
+ * @since 2.5.0
+ */
+// TODO once Java 8 is baseline move method to ChangeRequestSCMHead with default return value,
+// TODO deprecate this interface and add @Restricted(NoExternalUse.class) (retain empty interface for binary compat)
+public interface ChangeRequestSCMHead3 extends ChangeRequestSCMHead2 {
+
+    /**
+     * Returns the commit identifier (e.g. sha1) of the source branch.
```
jglick: Explain what a “source branch” would be. Give a concrete example using GitHub.

bitwiseman:
There definitely needs to be a JIRA for this, preferably one that links all the related changes together and summarizes the linked discussion thread into some cohesive design.

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @return the commit identifier of the source branch of this change request or @null if such a concept doesn't exists in the backing source control system.
+     */
+    @CheckForNull
+    String getSourceCommitId();
```
stephenc: The revision already contains this information. The head would change as the change request updates which is a antipattern

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.annotation.CheckForNull;
+
+/**
+ * Additional attributes of a {@link ChangeRequestSCMHead} that should have been in the original mixin but we are not
+ * targeting Java 8 so we cannot add the default methods to the interface and must have an ugly {@code 3} class instead.
+ * @since 2.5.0
+ */
+// TODO once Java 8 is baseline move method to ChangeRequestSCMHead with default return value,
+// TODO deprecate this interface and add @Restricted(NoExternalUse.class) (retain empty interface for binary compat)
+public interface ChangeRequestSCMHead3 extends ChangeRequestSCMHead2 {
```
stephenc: If we are changing, and I haven’t seen a compelling reason in this PR, then we should be moving methods back with the default modifier as we are on java 8 now.

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
henryju: How can I access the Revision from the `BranchNameContributor`? Is `projectFactory.getRevision(j)` the revision we want?

stephenc: That will return the revision the job last “discovered”

There may or may not be a build corresponding to that revision. It may be `null`

stephenc: But *why* should the name of the branch depend on the revision? That would change all the time. Users would object

henryju: `BranchNameContributor` is badly named IMO. It is in fact contributing all pipeline env variables related to multi-branch builds (normal branches, tags, pull requests). Some of those variables are already changing between each build (like `CHANGE_AUTHOR`)

stephenc: There's nothing stopping your own plugin from having its own name contributor... and that would be faster too, no need to wait for branch-api... and you can have complete control over the environment variable names

henryju: Still back to the point of how to get this properly:
* we could write our own `EnvironmentContributor`, but same as `BranchNameContributor`, what would be the correct way to get the SCMRevision that the about to be created build will refers to?
* we could instead put this logic in our [SimpleBuildWrapper](https://github.com/SonarSource/sonar-scanner-jenkins/blob/e0a26cecaf56671602cef60ddc84bf255283385a/src/main/java/hudson/plugins/sonar/SonarBuildWrapper.java#L93), doing something like:
    
```
public void setUp(Context context, Run<?, ?> build, FilePath workspace, Launcher launcher, TaskListener listener, EnvVars initialEnvironment)
    throws IOException, InterruptedException {
    SCMSource src = SCMSource.SourceByItem.findSource(build.getParent());
    SCMRevision revision = src != null ? SCMRevisionAction.getRevision(src, build) : null;
```

but then how to get the revision hash? I see that each implementation has its own way. For example in the [GitHub plugin](https://github.com/jenkinsci/github-branch-source-plugin/blob/e8469af7b665580dddfacc56825e023d6d9f43db/src/main/java/org/jenkinsci/plugins/github_branch_source/GitHubBuildStatusNotification.java#L82) or [Bitbucket plugin](https://github.com/jenkinsci/bitbucket-branch-source-plugin/blob/c8bbed32f1a7245c53753679f4c456667c774997/src/main/java/com/cloudbees/jenkins/plugins/bitbucket/BitbucketBuildStatusNotifications.java#L165)

We can't do that in our plugin without adding a dependency on each branch source impl (and I'm not sure their classes are intended to be stable).

Do you think it would be doable to expose this hash in the API, for example in ChangeRequestSCMRevision?

stephenc: `ChangeRequestSCMRevision.getTarget()` is the `SCMRevision` you are after for Git based source control you should be able to extract the revision from that as it should have a common Git baseclass...

stephenc: > ```
>     SCMSource src = SCMSource.SourceByItem.findSource(build.getParent());
>     SCMRevision revision = src != null ? SCMRevisionAction.getRevision(src, build) : null
> ```

Hmmm seems that there is no good reason why you need to look up the source to get the revision action... should be possible to get the SCMRevisionAction directly from the build.

stephenc: Build is actionable...

I think you can just do `build.getActions(SCMRevisionAction.class)` and then figure out which ones you are interested in...

(Hint: we suspect you still haven't noticed that a build can check out multiple repositories... or even multiple revisions of the same repository... and that might shake your "we need a single revision" world view)

stephenc:
Changes are unnecessary. A better path has been identified by using the SCM API's `SCMRevision` rather than attempting to mix revision information into the revision agnostic `SCMHead`

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
henryju: > ChangeRequestSCMRevision.getTarget() is the SCMRevision you are after

Isn't `ChangeRequestSCMRevision.getTarget()` the target of the PR? The javadoc of `ChangeRequestSCMHead::getTarget` says:
> Branch to which this change would be merged or applied if it were accepted

What I need is the source branch head commit revision.

> we suspect you still haven't noticed that a build can check out multiple repositories
I know, but similar to `GitHubBuildStatusNotification` and `BitbucketBuildStatusNotifications` we would like currently to focus on "simple" cases.

stephenc: So the revision itself should be that. I suspect what you really want is a Git-centric base interface of `ChangeRequestSCMRevision` in the Git plugin, e.g. `GitChangeRequestSCMRevision` and then get all the Git based SCM Sources to use that interface... either that or maintain special cases to extract from each of those plugin's custom types

simonvanderveldt:
> Changes are unnecessary. A better path has been identified by using the SCM API's `SCMRevision` rather than attempting to mix revision information into the revision agnostic `SCMHead`

@stephenc Where can this new approach be found? Is it already merged?

oxygenxo:
@stephenc to my best knowledge it isn't available without 'hacky' functions  yet.
Assuming my Jenkinsfile is cloned from PR's head I use the following function to get commit hash in my pipeline:
``` groovy
def commitHashForBuild(build) {
    def scmAction = build?.actions.find { action -> action instanceof jenkins.scm.api.SCMRevisionAction }
    if (scmAction?.revision instanceof org.jenkinsci.plugins.github_branch_source.PullRequestSCMRevision) {
        return scmAction?.revision?.pullHash
    } else if (scmAction?.revision instanceof jenkins.plugins.git.AbstractGitSCMSource$SCMRevisionImpl) {
        return scmAction?.revision?.hash
    } else {
        error("Build doesn't contain revision information. Do you run this from GitHub organization folder?")
    }
}
```
and then call it like this:
``` groovy
commit_sha = commitHashForBuild(currentBuild.rawBuild)
```

stephenc:
So write an extension plugin that provides a pipeline step to expose the info. My point is there is no need to modify the scm-api plugin. The api exposes all that is required to enable an extension plugin

simonvanderveldt:
Well, it seems rather like a bug to me that Jenkins/the git(hub) plugin(s) don't expose the correct git hash when merging a PR into the target branch.

Merging a branch into the target branch and reporting about the branch's latest commit are such basic functionalities that they shouldn't need workarounds/band aids.

oxygenxo:
Ah, I think I get it.
From plugin design perspective SCM API should not expose anything to builds, its main purpose is to provide, well, an API for other plugins, which can expose variables to jobs.
Sorry @stephenc , I'm just regular user and have no idea about plugins internals

stephenc:
It’s probably a design issue with the git plugin... otoh since the git plugin doesn’t actually have any concept of a pull request, it’s not really the git plugin’s fault.

Ideally the git plugin would provide PR base classes so that the child plugins stop re-inventing... but that should be a PR in the git plugin not a PR here.

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead3.java
timja: > Build is actionable...
> 
> I think you can just do `build.getActions(SCMRevisionAction.class)` and then figure out which ones you are interested in...
> 
> (Hint: we suspect you still haven't noticed that a build can check out multiple repositories... or even multiple revisions of the same repository... and that might shake your "we need a single revision" world view)

This doesn't seem to work at least in github-branch-source plugin, I added an EnvironmentContributor and that returns null

(all actions)
image

stephenc: Ahhh... so pipeline stores it differently, but it is stored just the same. You can get access to it from Branch API through the parent's helpers IIRC

timja: I'll try have another dig around later on, I wasn't able to find it in anything I had directly available, but I'm sure there's away

jcamerin:
Thank you @ababushk for the hack!  
Modified for Bitbucket
```
if (scmAction?.revision instanceof com.cloudbees.jenkins.plugins.bitbucket.PullRequestSCMRevision) {
    return scmAction?.revision?.pull.hash
}
```

sparrowt:
A simpler alternative is `pullRequest.head` assuming you have the [Pipeline: GitHub](https://plugins.jenkins.io/pipeline-github/) plugin installed, which provides a global called `pullRequest` which is available if you are in a PR job e.g. if `env.CHANGE_ID`

This gets the actual head SHA of the PR, rather than `GIT_COMMIT` which is the temporary merge commit with master/main.

See: https://github.com/jenkinsci/pipeline-github-plugin/blob/master/README.md#pullrequest

PayBas:
A damn shame this was never implemented.

Now we keep having to resort to workarounds like:
```
source = sh(returnStdout: true, script: "git rev-parse origin/${env.CHANGE_BRANCH}").trim()
```

PR #73, opened 2019-07-11
Push minimum jenkins version back to 2.107.3

bitwiseman:


jglick:
> #61 for a backlink

Unfortunately does not currently work from a PR review comment, as I think I have reported to @github. But this should. :smile:

PR #74, opened 2019-07-11
[JENKINS-43802] SCMSource addenda

jglick:
Amendments to #65.

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -628,7 +628,7 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      * {@code null} if malformed or not found
      * @throws IOException if an error occurs while performing the operation.
      * @throws InterruptedException if any thread has interrupted the current thread.
-     * @since 2.5
+     * @since 2.6.0
      */
     @CheckForNull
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -705,12 +705,12 @@ protected SCMRevision retrieve(@NonNull final String thingName, @NonNull TaskLis
...
      * @throws IOException if an error occurs while performing the operation.
      * @throws InterruptedException if any thread has interrupted the current thread.
-     * @since 2.5
+     * @since 2.6.0
      */
     @NonNull
     public final Set<String> fetchRevisions(@CheckForNull TaskListener listener, @CheckForNull Item context)
             throws IOException, InterruptedException {
-        return retrieveRevisions(defaultListener(listener));
+        return retrieveRevisions(defaultListener(listener), context);
```
jglick: Not as bad as it sounds since the only proposed call site is [currently commented out anyway](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/8dae8d968b2ec8ea727c83518280c6e0ee3d51fd/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java#L184-L192).

PR #75, opened 2019-08-01
Improve event support impl doc

baymac:
I am currently working on `GitLab Branch Source Plugin`. I faced some difficulties with implementing web hooks support and I would like to suggest some improvements. Please let me know what you think about this.

### DIFF
--- a/docs/implementation.adoc
+++ b/docs/implementation.adoc
@@ -1353,11 +1353,12 @@ The first part of implementing event support is to determine how events will be 
 There are a number of techniques that can be used. The two most common techniques are:
 
 * Webhook
-* Messaging service
 
 The webhook technique typically involves setting up a `RootAction` that can receive a payload from the source control system.
 For this technique to work, the source control system must be able to establish a connection to the Jenkins server.
 This can be problematic where, for example, the Jenkins server is on an internal-only network and the source control system is an externally hosted service (e.g. GitHub)
+
+* Messaging service
 
 The messaging service uses a broker which can be reached by both the Jenkins server and the source control system.
 The source control system sends its event payloads to the broker system.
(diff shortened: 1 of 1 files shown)


docs/implementation.adoc
```diff
@@ -1353,26 +1353,30 @@ The first part of implementing event support is to determine how events will be
...
+
 The basic starting point for a WebHook is an `UnprotectedRootAction`
 
 [source,java]
 ----
 @Extension
 public class MySCMWebHook implements UnprotectedRootAction {
-    private static final String URL_NAME = "my-scm-hook";
+    public static final String URL_NAME = "my-scm-hook";
+    public static final String ENPOINT = "notify";
```
bitwiseman:
```suggestion
    public static final String ENDPOINT = "notify";
```

docs/implementation.adoc
```diff
@@ -1428,6 +1432,8 @@ public class MySCMWebHook implements UnprotectedRootAction {
 }
 ----
 
+Basically, you don't have to worry about triggering the build... you just have to fire the event. If you event is implemented correctly then the event subsystem will match the event to the corresponding SCMSource instances and let them do what they need to do (and in the case of multibranch, what it needs to do is: initiate a check of the latest revision and then build if the revision is changed).
```
bitwiseman: Started with typo fix then added a bit more. 
```suggestion
You don't have to worry about triggering the build, you just have to fire the event. If your event is implemented correctly then the event subsystem will match the event to the corresponding SCMSource instances and let them do what they need to do. For example, in the case of multibranch, they will initiate a check of the latest revision and then build if the revision is changed).
```

bitwiseman: @baymac  Can you accept this suggestion?  Then we can merge this.

PR #79, opened 2019-10-23
Custom Avatar sources for AvatarCache

mlasevich:
As per https://issues.jenkins-ci.org/browse/JENKINS-59797

While trying to fix a problem with a plugin relying on this one, I have identified that AvatarCache only works for simple http/https urls without authentication, which obviously does not work for many SCMs that require authentication. In order to make it more flexible, I have added ability for SCM implementations to provide their own image sources and methods for fetching them instead of relying on what is hardcoded into AvatarCache.

I have included one implementation which copies the existing functionality and wrapped old methods using just URL to using that, so that this should continue to work with all existing code.

### DIFF
--- a/src/main/java/jenkins/scm/impl/avatars/AvatarCache.java
+++ b/src/main/java/jenkins/scm/impl/avatars/AvatarCache.java
@@ -37,13 +37,9 @@ import hudson.util.NamingThreadFactory;
 import java.awt.Color;
 import java.awt.Graphics2D;
 import java.awt.image.BufferedImage;
-import java.io.BufferedInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
-import java.net.HttpURLConnection;
-import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
@@ -65,7 +61,8 @@ import javax.imageio.ImageIO;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletResponse;
 import jenkins.model.Jenkins;
-import org.apache.commons.io.IOUtils;
+import jenkins.scm.impl.avatars.AvatarCacheSource.AvatarImage;
+
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.QueryParameter;
@@ -88,6 +85,11 @@ public class AvatarCache implements UnprotectedRootAction {
      * Our logger.
      */
     private static final Logger LOGGER = Logger.getLogger(AvatarCache.class.getName());
+    /**
+     * URI For this action
+     */
+    private static final String ActionURI = "avatar-cache";
+
     /**
      * Maximum concurrent requests to fetch images.
      */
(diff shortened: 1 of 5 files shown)


mlasevich:
*poke* - How do we get this reviewed?

bitwiseman:
@mlasevich 
I can take a look.

src/main/java/jenkins/scm/impl/avatars/AvatarCache.java
```diff
@@ -580,59 +603,23 @@ public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object nod
      * A task to fetch an image from a remote URL.
      */
     private static class FetchImage implements Callable<CacheEntry> {
-        private final String url;
+        private final AvatarCacheSource source;
```
bitwiseman: It seems really odd to me to have this vanishingly small class...   Could this be made into part of the `AvatarCacheSource`?  Should it?

src/main/java/jenkins/scm/impl/avatars/AvatarCacheSource.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.awt.image.BufferedImage;
+
+/**
+ *
+ * Interface for Avatar Cache Item Source
+ * 
+ * This defines a source for avatar to be ached and implementation to fetch it
+ */
+public interface AvatarCacheSource {
```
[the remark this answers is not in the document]
mlasevich: Fair enough. I was thinking of it in terms of it being a Source for AvatarCache service, but it does represent an image, so it is probably more logical your way.

src/main/java/jenkins/scm/impl/avatars/AvatarCacheSource.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Holds Image and lastModified date
+     */
+    public static class AvatarImage {
```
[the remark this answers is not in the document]
mlasevich: Part laziness, part trying to keep it obvious that it is part of the AvatarCacheSource.... but mostly laziness

src/main/java/jenkins/scm/impl/avatars/AvatarCache.java
mlasevich: Hmm. I see what you are saying, but after thinking about it, I think it may make sense to leave it alone. `AvatarCacheSource` is an interface intended to be implemented by other plugins - and this functionality is an internal part of AvatarCache, that could possibly change in the future. We could, of course, provide an abstract class instead of an interface, but my gut feeling is we should keep it simple and not mix the two.

src/main/java/jenkins/scm/impl/avatars/AvatarCacheSource.java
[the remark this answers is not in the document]
mlasevich: Renamed

src/main/java/jenkins/scm/impl/avatars/AvatarCacheSource.java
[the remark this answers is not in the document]
mlasevich: broke it out to a separate class

mlasevich:
Ok, I took care of most of your comments, going to take a stab at unit tests

mlasevich:
After looking at the unit tests, it occurs to me that the existing tests already cover new functionality as old calls call new calls now.

src/main/java/jenkins/scm/impl/avatars/UrlAvatarCacheSource.java
```diff
@@ -0,0 +1,137 @@
...
+                    new Object[] { url, duration });
+        }
+        return AvatarImage.EMPTY;
+    }
+
+    /**
+     * Generate hash key used for caching
+     */
+    @Override
+    public String hashKey() {
```
bitwiseman:
```suggestion
    public String getId() {
```

Sorry, I thought I'd said this in the first review.  This is not a hash, it is an ID string.  If we get a new items with the same hash it might just be a hash collision. If we get a new item with the same id, it should always replace the old item, right?

bitwiseman: @mlasevich 
https://github.com/mlasevich/scm-api-plugin/pull/1

mlasevich: fair enough. :-)

src/main/java/jenkins/scm/impl/avatars/AvatarCache.java
mlasevich: I'll let you resolve this one, if you are satisfied :-)

mlasevich:
Is there anything else you are waiting on me for?

mlasevich:
Thanks. Is there an eta for it to be released? I would like to update the bitbucket branch source plugin this was originally written for to use this instead of its own copy :-)

mlasevich:
Any idea when this (2.6.4) may be released?

jetersen:
@bitwiseman Liam is there a chance for a release 😅

PR #80, opened 2020-01-10
Use GitHub for plugins.jenkins.io docs

MarkEWaite:
The consumer docs in the plugin are great. We should use them on the plugin site instead of referring to the wiki page.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
   <description>
     This plugin provides a new enhanced API for interacting with SCM systems.
   </description>
-  <url>https://github.com/jenkinsci/scm-api-plugin/master/docs/consumer.adoc</url>
+  <url>https://github.com/jenkinsci/scm-api-plugin/blob/master/docs/consumer.adoc</url>
   <licenses>
     <license>
       <name>The MIT license</name>
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
   <description>
     This plugin provides a new enhanced API for interacting with SCM systems.
   </description>
-  <url>https://wiki.jenkins.io/display/JENKINS/SCM+API+Plugin</url>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -41,7 +41,7 @@
   <description>
     This plugin provides a new enhanced API for interacting with SCM systems.
   </description>
-  <url>https://wiki.jenkins.io/display/JENKINS/SCM+API+Plugin</url>
+  <url>https://github.com/jenkinsci/scm-api-plugin/master/docs/consumer.adoc</url>
```
MarkEWaite: I think I've made a mistake here.  I need to double check the instructions to confirm this is the correct way to refer to an adoc file in the GitHub repository. I suspect an error in the URL.

timja: you need to add tree to the URL I believe

MarkEWaite:
```suggestion
  <url>https://github.com/jenkinsci/scm-api-plugin/blob/master/docs/consumer.adoc</url>
```

MarkEWaite:
> Also need to update the README and pull in the changelog from the wiki.

Done.  All references to wiki have been replaced with references to plugins.jenkins.io.  CHANGELOG.md has been created.

Did not attempt to modify the plugin to use release drafter.  Intentionally out of scope of this pull request.

MarkEWaite:
Fixing the CI build failures will require pom changes.  Being based on pom 3.47 and supporting Jenkins 2.107 all the way to 2.222.3 is too wide a range to not require pom changes.

The build failures are due to the recent change in pipeline recommended configuration.  It now tests 2.222.3 as the "high end" Jenkins build version

timja:
github was a bit keen on this one, the magic fix word shut it by accident, after merging master the build should pass though

PR #87, opened 2021-03-16
Bump baseline, remove pointless test breaking CI

timja:
Relates to https://github.com/jenkinsci/branch-api-plugin/pull/242, blocks https://github.com/jenkinsci/scm-api-plugin/pull/85

Messages are excluded from coverage in the parent pom: https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml#L1081

This broke when coverage was enabled by default, I dug into it a bit last time but couldn't see a fix other than removing it, which seems correct anyway


timja:
@dwnusbaum

dwnusbaum (approved):

bitwiseman:
@timja I think you pulled these changes into #85, right?

car-roll (approved):
Looks fine outside of what @bitwiseman said, the `MessagesTest` classes appear to already be removed

timja:
Yes in case anyone had issues with the other PR this one could have been merged first but all good

PR #101, opened 2021-11-15
Inline usages of Guava

basil:
Eliminates any possible exposure to bugs like [JENKINS-67135](https://issues.jenkins.io/browse/JENKINS-67135) by removing any usages of Guava from this plugin. In their place, we inline the relevant Guava code, with some minor adjustments and preserving its license.

How to review:

- First, review the changes to `src/main/java/jenkins/scm/api/SCMName.java`.
- Then, review the changes to the inlined Guava classes compared to the originals. For convenience, I've put up [a diff here](https://github.com/basil/scm-api-plugin/commit/83b18738bf196abf5a81874299a79abf83a334ef). These changes are mostly removing code that we don't need and adjusting various code to use the Java Platform instead of Guava to avoid having to inline even more classes.


basil:
IIRC @timja had been working on a change like this in #85.

timja:
Why not just bump the baseline to 2.320?

basil:
> Why not just bump the baseline to 2.320?

I'm not sure that would accomplish the stated objective (eliminating any possible exposure to bugs like [JENKINS-67135](https://issues.jenkins.io/browse/JENKINS-67135)). But I suspect this PR would accomplish that objective by eliminating any references to the problematic class described in the bug.

olamy:
this is adding a lot of code to maintain here. security issues to fix.
are those Guava classes really a problem when upgrading?

basil:
[JENKINS-67135](https://issues.jenkins.io/browse/JENKINS-67135) was closed, so I think this can be closed as well. Glad it wasn't needed!

PR #107, opened 2022-01-07
Update parent, baseline, BOM

jglick:
Supersedes #103.


pom.xml
```diff
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.6.6</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.235.5</jenkins.version>
+    <jenkins.version>2.289.1</jenkins.version>
```
jtnord, line 70: this version seems a bit arbirary?  it is the oldest cloudbees supports, not the newest LTS line, and it is not clear that anything here is needed from this version that is not in older versions, nor why a newer versions was not picked.

jtnord (approved):

pom.xml
jglick: 2.249.x is the minimum as of https://github.com/jenkinsci/plugin-pom/pull/480. 2.289.x allows use of current `bom` releases as described in https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.32.

PR #108, opened 2022-01-07
Enable CD

jglick:
https://www.jenkins.io/doc/developer/publishing/releasing-cd/ https://github.com/jenkins-infra/repository-permissions-updater/pull/2304


MarkEWaite (approved):

pom.xml
```diff
@@ -65,8 +55,7 @@
   </scm>
 
   <properties>
-    <revision>2.6.6</revision>
-    <changelist>-SNAPSHOT</changelist>
+    <changelist>999999-SNAPSHOT</changelist>
```
jtnord, line 58: personally I really dislike this format, especially when it pertains to an API plugin as it looses any possibility of a semver like headup (this is a major version bump)

jglick: This plugin historically used major.minor.micro version format. In practice no one consistently paid attention to the choice of version used for various PRs, so we likely had any number of API additions or other features shipped in supposedly micro releases. As far as incompatible changes, we had one total rewrite five years ago, for which the 2.0.0 may have been a helpful mnemonic for plugin developers, but not particularly for users since the Jenkins plugin manager will simply offer it as an update like any other. If you _know in advance_ you are going to make an incompatible change, signalling this in a version number is barely helpful—you anyway need to find all API clients or implementations which would be broken, and prepare them. Most incompatible changes are accidental regressions, for which the version number would come too late. So I do not find the Semver style helpful for Jenkins plugins. For some Apache library that gets released once every few months and which people consciously decide to upgrade or not, sure.

PR #124, opened 2022-03-25
JENKINS-68116 Make event thread pool size customize-able

timja:
See https://issues.jenkins.io/browse/JENKINS-68116

Slow processing of events is causing delays of over an hour for our users to have their builds / pushes picked up

Currently increases thread pool size from 10 -> 30 and adds a system property to make it customise-able

Last changed in https://github.com/jenkinsci/scm-api-plugin/pull/61

I've ran this for four hours on a busy controller and never saw the queue grow above 4 or the thread pool get maxed out with running threads (highest I saw was ~27).

Running this in script console shows the status:
```
$ jenkins.scm.api.SCMEvent.executorService

Result: java.util.concurrent.ScheduledThreadPoolExecutor@23c09aa1[Running, pool size = 30, active threads = 20, queued tasks = 0, completed tasks = 4029]
```

Related PR: https://github.com/jenkinsci/branch-api-plugin/pull/304

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMEvent.java
+++ b/src/main/java/jenkins/scm/api/SCMEvent.java
@@ -46,6 +46,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.servlet.http.HttpServletRequest;
 import jenkins.security.ImpersonatingScheduledExecutorService;
+import jenkins.util.SystemProperties;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -106,6 +107,9 @@
      * @since 2.0.3
      */
     public static final String ORIGIN_UNKNOWN = "?";
+
+    private static final int EVENT_THREAD_POOL_SIZE = SystemProperties
+        .getInteger(SCMEvent.class.getName() + ".EVENT_THREAD_POOL_SIZE", 30);
```
timja, line 112: debating whether or not to increase this or leave it at the previous value of 10, was also wondering about dynamically increasing the thread pool size if the queue get's too big.

timja:
I'm going to do some more log searching now that I've increased the amount of logs to see if there's anything useful.
I'm wondering if we have some repositories with a lot of branches that take awhile to process.

car-roll (approved):

olamy (approved):

timja:
ping @jtnord

timja:
Ping @jtnord / @jenkinsci/scm-api-plugin-developers

I've reduced our folder count from 68 to 6 which seems to be performing better but I would still like to be able to configure this above 10. It doesn't make sense for a 10 repository instance to have the same processing capacity as an org with over 1500 repositories

jtnord (approved):

PR #127, opened 2022-04-04
 JENKINS-68116 Export metrics for scm event queue performance

timja:
I would like to be able to monitor our SCM event queue and be able to measure the effect that tuning I do has on it.

see https://issues.jenkins.io/browse/JENKINS-68116

Implemented in the open telemetry plugin https://github.com/jenkinsci/opentelemetry-plugin/pull/388

image

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMEvent.java
+++ b/src/main/java/jenkins/scm/api/SCMEvent.java
@@ -135,6 +135,7 @@ public abstract class SCMEvent<P> {
      * The scheduled executor thread pool. This is initialized lazily since it may be never needed.
      */
     private static ScheduledExecutorService executorService;
+    private static ScheduledThreadPoolExecutor threadPoolExecutor;
 
     /**
      * Constructor to use when the timestamp is available from the external SCM.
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/scm/api/SCMEvent.java
```diff
@@ -210,13 +211,41 @@ protected SCMEvent(SCMEvent<P> copy) {
...
         return executorService;
     }
 
+    public static EventQueueMetrics getEventProcessingMetrics() {
+        return new EventQueueMetrics();
+    }
+
+    public static class EventQueueMetrics {
+        public int getPoolSize() {
+            return threadPoolExecutor == null ? 0 : threadPoolExecutor.getPoolSize();
```
timja, line 233: lazily initialised and calculated as required due to otel treating each resource as completely independent, originally I set all the fields in the constructor but that required fetching the data a lot more than required (some of the calls acquire locks)

timja:
cc @jtnord / @car-roll could you take a look please ❤️

timja:
ping @jtnord / @car-roll

jtnord (approved):

jtnord:
would seem like a reasonable addition to make to the `metrics` plugin also (then it would be visible in datadog,prometheus etc (anything that can use JMX)

jtnord (approved):

timja:
> would seem like a reasonable addition to make to the `metrics` plugin also (then it would be visible in datadog,prometheus etc (anything that can use JMX)

With open telemetry it trivial to send to Prometheus, and from a quick read Datadog as well

PR #140, opened 2022-05-07
Make a comment more inclusive

Ladyprowess:
Rather than "master" use "controller"

~~Ensure you have provided tests - that demonstrate feature works or fixes the issue~~

### DIFF
--- a/docs/consumer.adoc
+++ b/docs/consumer.adoc
@@ -63,7 +63,7 @@ Jenkins then notifies you and the developer of the other feature branch that the
 
 In order to allow for some rollback of some of these side-effects, typically one would:
 
-* Create a job that just builds the master branch.
+* Create a job that just builds the controller branch.
 * Create one job each for each of the sustaining branches that need to be tracked.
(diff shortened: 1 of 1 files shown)


docs/consumer.adoc
```diff
@@ -63,7 +63,7 @@ Jenkins then notifies you and the developer of the other feature branch that the
 
 In order to allow for some rollback of some of these side-effects, typically one would:
 
-* Create a job that just builds the master branch.
+* Create a job that just builds the controller branch.
```
aHenryJard, line 66: This "master" word seems to be about git branches, so "controller" does not works here, if I remember correctly the "main branch" can be used.

```suggestion
* Create a job that just builds the main branch.
```

MarkEWaite: I agree with the suggestion from @aHenryJard .  The "main" word is preferred here.

MarkEWaite (approved):

car-roll (approved):

PR #78, opened 2019-10-14
[JENKINS-48431] Support both lightweight checkout AND build parameters

tgatinea:
[JENKINS-48431](https://issues.jenkins-ci.org/browse/JENKINS-48431) To support both lightweight checkout AND build parameters, we need to:

- recover the environment parameters of the build
- evaluate the SCM structure according to the build environment parameters

This leads to implement a new method :
```
public static SCMFileSystem of(@nonnull Run build, @nonnull SCM scm, @checkfornull SCMRevision rev)
```

This new method offer the capacity to recover the environments parameter of the build as the current build is propagated through the new method.

Note that this modification is in relation with two other pull requests that concerns:
- workflow-cps-plugin: https://github.com/jenkinsci/workflow-cps-plugin/pull/329
- git-plugin: https://github.com/jenkinsci/git-plugin/pull/772

PR #160, opened 2022-07-28
New API for SCMFileSystem to allow for lightweight checkout with build params

MartinKosicky:
I added a new Builder class into the SCMFileSystem so that we have backward compatibility with current SCM implementations and with ability for a new implementation, so that they can make for example branch substitutions based on parameters in the build (environment).


src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -527,4 +556,27 @@ public SCMFileSystem build(@NonNull SCMSource source, @NonNull SCMHead head,
...
+         * @param owner the owner of the {@link SCM}
+         * @param scm the {@link SCM}.
+         * @param rev the specified {@link SCMRevision}.
+         * @return the corresponding {@link SCMFileSystem} or {@code null} if this builder cannot create a {@link
+         * SCMFileSystem} for the specified {@link SCM}.
+         * @throws IOException          if the attempt to create a {@link SCMFileSystem} failed due to an IO error
+         *                              (such as the remote system being unavailable)
+         * @throws InterruptedException if the attempt to create a {@link SCMFileSystem} was interrupted.
+         */
+        @CheckForNull
```
MartinKosicky, line 577: in GitSCM I can make GitSCMFileSystem to override from this builder instead of the old and pass the Run<?,?>

MartinKosicky:
I would fix the blockers ASAP but first I would like some feedback if this concept does make sence, or if you would prefer a different approach. I dont have problems doing it, but I dont want to do major fixes if I am not sure I am on the right track :)

MartinKosicky:
I checked it working by updating also the , I also have ready in local workflow-cps plugin and done some local changes to git plugin, that can be only added for PR if this PR will be succesful

jglick:
Did you look at #78?

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -527,4 +556,29 @@ public SCMFileSystem build(@NonNull SCMSource source, @NonNull SCMHead head,
             return build(owner, source.build(head, rev), rev);
         }
     }
+
+    public abstract static class Builder2 extends Builder {
```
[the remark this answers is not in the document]
MartinKosicky: Oki, well I added a new overridable method, instead of a new class with abstract method. Although I couldn't find a suitable place where to use the Util.isOverridden,  if the target library now doesnt implement the new method, it will be automatically called the base method that will delegate to old method. Is that OK ?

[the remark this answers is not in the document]
jglick: OK if all calls to `build` go through `SCMFileSystem.of`, which may be true. Still make the original method not be `abstract` (have it throw `AbstractMethodError`) so an impl need only override the new overload.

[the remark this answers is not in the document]
MartinKosicky: Ok I removed the abstract modifier and added a throw with notice to implement the new version. Also changed the other call to build that should be calling the new implementation.  I will try to figure out how to make the draft PR in the workflow-csp and git plugin once i figure out how this incremental feature works

MartinKosicky:
I added draft https://github.com/jenkinsci/workflow-cps-plugin/pull/577  and https://github.com/jenkinsci/git-plugin/pull/1305

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -187,15 +187,37 @@ public static SCMFileSystem of(@NonNull Item owner, @NonNull SCM scm) throws IOE
      * @throws InterruptedException if the attempt to create a {@link SCMFileSystem} was interrupted.
      */
     @CheckForNull
-    public static SCMFileSystem of(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev)
+    public static SCMFileSystem of(@NonNull Item owner, @NonNull SCM scm,@CheckForNull SCMRevision rev) throws IOException, InterruptedException {
+        return of(owner, scm, rev,null);
```
jglick, line 191:
```suggestion
    public static SCMFileSystem of(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev) throws IOException, InterruptedException {
        return of(owner, scm, rev, null);
```

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -187,15 +187,37 @@ public static SCMFileSystem of(@NonNull Item owner, @NonNull SCM scm) throws IOE
...
+
+
+    /**
+     * Given a {@link SCM} this method will try to retrieve a corresponding {@link SCMFileSystem} instance that
+     * reflects the content at the specified {@link SCMRevision}.
+     *
+     * @param owner the owner of the {@link SCM}
+     * @param scm the {@link SCM}.
+     * @param rev the specified {@link SCMRevision}.
+     * @param build the specified {@link Run}.
```
jglick, line 202: None of these tags actually tell the reader anything. The existing ones are not your doing, but while we are adding new ones, suggest perhaps
```suggestion
     * @param build an associated build context, if any, that could be used for example to look up parameters
```

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -500,8 +522,11 @@ public final boolean supports(SCMSourceDescriptor descriptor) {
          * @throws InterruptedException if the attempt to create a {@link SCMFileSystem} was interrupted.
          */
         @CheckForNull
-        public abstract SCMFileSystem build(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev)
-                throws IOException, InterruptedException;
+        public SCMFileSystem build(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev)
+                throws IOException, InterruptedException
+        {
+            throw new AbstractMethodError("Implement build(Item, SCM, SCMRevision,Run<?,?>)");
```
jglick, line 528:
```suggestion
            throw new AbstractMethodError("Implement build(Item, SCM, SCMRevision, Run)");
```

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -525,12 +550,38 @@ public SCMFileSystem build(@NonNull SCMSource source, @NonNull SCMHead head,
...
+         *                              (such as the remote system being unavailable)
+         * @throws InterruptedException if the attempt to create a {@link SCMFileSystem} was interrupted.
+         */
+        @CheckForNull
+        public SCMFileSystem build(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev,
+                                            @CheckForNull Run<?,?> _build)
+                throws IOException, InterruptedException
+        {
+            // if this is not overriden, fallback to the previous implementation
+            return build(owner,scm, rev);
```
jglick, line 584:
```suggestion
            return build(owner, scm, rev);
```

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -525,12 +550,38 @@ public SCMFileSystem build(@NonNull SCMSource source, @NonNull SCMHead head,
...
+         * @throws IOException          if the attempt to create a {@link SCMFileSystem} failed due to an IO error
+         *                              (such as the remote system being unavailable)
+         * @throws InterruptedException if the attempt to create a {@link SCMFileSystem} was interrupted.
+         */
+        @CheckForNull
+        public SCMFileSystem build(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev,
+                                            @CheckForNull Run<?,?> _build)
+                throws IOException, InterruptedException
+        {
+            // if this is not overriden, fallback to the previous implementation
```
jglick, line 583:
```suggestion
            // if this is not overridden, fall back to the previous implementation
```

jglick (approved):

src/main/java/jenkins/scm/api/SCMFileSystem.java
MartinKosicky: Hmm but that wouldnt match the signature?  . And I made the signature that way because I saw it also in GitSCM.checkout

jglick: Generic types are not part of the method signature at the bytecode level.

MartinKosicky:
Hello @jglick , I would just like to ask if the current flow is that multiple reviewers have to approve or if there are any improvements I ca do here regarding this scm-api :)

jglick:
> if the current flow is that multiple reviewers have to approve

Not in general, though for significant changes and especially API additions (which cannot generally be reverted if they turn out to be misconceived) I prefer to have a second opinion.

MartinKosicky:
> > if the current flow is that multiple reviewers have to approve
> 
> Not in general, though for significant changes and especially API additions (which cannot generally be reverted if they turn out to be misconceived) I prefer to have a second opinion.

Yes that sounds reasonable. I will wait :)

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -525,12 +550,38 @@ public SCMFileSystem build(@NonNull SCMSource source, @NonNull SCMHead head,
...
             Class<?> enclosingClass = getClass().getEnclosingClass();
             return enclosingClass != null && descriptor.clazz.isAssignableFrom(enclosingClass);
         }
+
+        /**
+         * Given a {@link SCM} this should try to build a corresponding {@link SCMFileSystem} instance that
+         * reflects the content at the specified {@link SCMRevision}. If the {@link SCM} is supported but not
+         * for a fixed revision, best effort is acceptable as the most capable {@link SCMFileSystem} will be returned
+         * to the caller. If the {@link Run} is provided, it can be used to alter the behaviour of buil based on the current Run
+         * for example, expand branch name based on current build properties, etc.
```
dwnusbaum, line 566: There is a typo here `buil` -> `build`, but I might reword things a bit:

```suggestion
         * to the caller. If the {@link Run} is provided, it can be used to alter the behavior. 
         * For example, variables in the branch name can be expanded based on current build properties, etc.
```

src/main/java/jenkins/scm/api/SCMFileSystem.java
```diff
@@ -500,8 +522,11 @@ public final boolean supports(SCMSourceDescriptor descriptor) {
          * @throws InterruptedException if the attempt to create a {@link SCMFileSystem} was interrupted.
          */
         @CheckForNull
-        public abstract SCMFileSystem build(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev)
-                throws IOException, InterruptedException;
+        public SCMFileSystem build(@NonNull Item owner, @NonNull SCM scm, @CheckForNull SCMRevision rev)
+                throws IOException, InterruptedException
+        {
+            throw new AbstractMethodError("Implement build(Item, SCM, SCMRevision, Run)");
```
dwnusbaum, line 528: nit: Maybe this would already be clear to developers, but this API may still be called directly and so both overloads need to be implemented.

```suggestion
            throw new AbstractMethodError("Implement build(Item, SCM, SCMRevision, Run) and override this method to delegate to it");
```

dwnusbaum (approved):
The new APIs seem fine to me.

car-roll (approved):

PR #180, opened 2022-12-15
[JENKINS-46795] `TrustworthyBuild` extension point

jglick:
https://github.com/jenkinsci/workflow-multibranch-plugin/pull/220#discussion_r1049575626

### DIFF
--- a/src/main/java/jenkins/scm/api/SCMSource.java
+++ b/src/main/java/jenkins/scm/api/SCMSource.java
@@ -77,6 +77,8 @@ import org.kohsuke.stapler.export.ExportedBean;
 public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
         implements ExtensionPoint {
 
+    private static final Logger LOGGER = Logger.getLogger(SCMSource.class.getName());
+
     /**
      * Replaceable pronoun of that points to a {@link SCMSource}. Defaults to {@code null} depending on the context.
      * @since 2.0
@@ -956,11 +958,35 @@ public abstract class SCMSource extends AbstractDescribableImpl<SCMSource>
      * @throws IOException in case the implementation must call {@link #fetch(SCMHead, TaskListener)} or similar
      * @throws InterruptedException in case the implementation must call {@link #fetch(SCMHead, TaskListener)} or similar
      * @since 1.1
+     * @see #getTrustedRevisionForBuild
      */
     @NonNull
     public SCMRevision getTrustedRevision(@NonNull SCMRevision revision, @NonNull TaskListener listener)
             throws IOException, InterruptedException {
         return revision;
+    }
+
+    /**
+     * Refined version of {@link #getTrustedRevision(SCMRevision, TaskListener)} that takes into account the build context.
+     * @param build a running build
+     * @return {@link #getTrustedRevision} if the build itself does not indicate trust;
+     *         simply {@code revision} if any {@link TrustworthyBuild} says that it does
+     */
+    @NonNull
+    public final SCMRevision getTrustedRevisionForBuild(@NonNull SCMRevision revision, @NonNull TaskListener listener, @NonNull Run<?, ?> build)
+            throws IOException, InterruptedException {
+        if (ExtensionList.lookup(TrustworthyBuild.class).stream().anyMatch(tb -> tb.shouldBeTrusted(build))) {
+            LOGGER.fine(() -> build + " with " + build.getCauses() + " was considered trustworthy, so using " + revision + " as is");
+            return revision;
+        } else {
+            SCMRevision trustedRevision = getTrustedRevision(revision, listener);
+            if (trustedRevision.equals(revision)) {
+                LOGGER.fine(() -> build + " was not considered trustworthy, but " + revision + " was trusted anyway");
+            } else {
+                LOGGER.fine(() -> build + " was not considered trustworthy, so replacing " + revision + " with " + trustedRevision);
+            }
+            return trustedRevision;
+        }
     }
 
     /**
--- a/src/main/java/jenkins/scm/api/TrustworthyBuild.java
+++ b/src/main/java/jenkins/scm/api/TrustworthyBuild.java
@@ -1,0 +1,68 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2022 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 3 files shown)


src/main/java/jenkins/scm/impl/TrustworthyBuilds.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class TrustworthyBuilds {
+
+    // Also effectively handles ReplayCause since that is only ever added in conjunction with UserIdCause. (see ReplayAction.run2)
+    @Extension
+    public static TrustworthyBuild byUserId() {
+        return TrustworthyBuild.byCause(Cause.UserIdCause.class);
+    }
+
+    // TODO until github-checks can declare a dep on a sufficiently new scm-api
```
jglick, line 39: https://github.com/jenkinsci/github-checks-plugin/pull/307

src/main/java/jenkins/scm/api/TrustworthyBuild.java
```diff
@@ -0,0 +1,68 @@
...
+     * Should this build be trusted to load sensitive source files?
+     * If any implementation returns true then it is trusted.
+     */
+    boolean shouldBeTrusted(@NonNull Run<?, ?> build);
+
+    /**
+     * Convenience for the common case that a particular trigger cause indicates trust.
+     * Examples of causes which could be trusted include:
+     * <ul>
+     * <li>{@link Cause.UserIdCause}
```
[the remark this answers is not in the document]
jglick: :thinking: you are probably right. Would be better to check not only that `UserIdCause` is present but that the named user in fact has `Item.CONFIGURE`. That locks out people with only `ReplayAction.REPLAY` but I think that is normally just implied by `CONFIGURE`, and anyway this would just restore the status quo basically (except failing the build rather than silently proceeding with the stock `Jenkinsfile`).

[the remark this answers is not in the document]
jglick: Working on it. The problem I hit is that for a branch project, `CONFIGURE` is unconditionally denied (whereas `REPLAY` is granted).

src/main/java/jenkins/scm/impl/TrustworthyBuilds.java
```diff
@@ -0,0 +1,84 @@
...
+                return false;
+            }
+            var user = User.getById(userId, false);
+            if (user == null) {
+                listener.getLogger().printf("Not trusting build since no user ‘%s’ is known%n", userId);
+                return false;
+            }
+            try {
+                var permission = Run.PERMISSIONS.find("Replay"); // ReplayAction.REPLAY
+                if (permission == null) { // no workflow-cps
```
jglick, line 58: _Could_ be defined more modularly as an extension in `workflow-cps`, but would be awkward since the rest of this impl would need to be duplicated, and `ordinal` used to avoid printing misleading messages.

In practice the callers of `getTrustedRevisionForBuild` are going to be in Pipeline code so I am not too concerned.

src/main/java/jenkins/scm/impl/TrustworthyBuilds.java
```diff
@@ -0,0 +1,84 @@
...
+            if (cause == null) {
+                // probably some other cause; do not print anything
+                return false;
+            }
+            var userId = cause.getUserId();
+            if (userId == null) {
+                listener.getLogger().println("Not trusting build since no user name was recorded");
+                return false;
+            }
+            var user = User.getById(userId, false);
```
jglick, line 51: Could pass `true` but we expect the `User` to have just been loaded anyway.

src/main/java/jenkins/scm/api/SCMSource.java
```diff
@@ -956,13 +958,37 @@ public final SCM build(@NonNull SCMHead head) {
...
+    @NonNull
+    public final SCMRevision getTrustedRevisionForBuild(@NonNull SCMRevision revision, @NonNull TaskListener listener, @NonNull Run<?, ?> build)
+            throws IOException, InterruptedException {
+        if (ExtensionList.lookup(TrustworthyBuild.class).stream().anyMatch(tb -> tb.shouldBeTrusted(build, listener))) {
+            LOGGER.fine(() -> build + " with " + build.getCauses() + " was considered trustworthy, so using " + revision + " as is");
+            return revision;
+        } else {
+            SCMRevision trustedRevision = getTrustedRevision(revision, listener);
+            if (trustedRevision.equals(revision)) {
+                LOGGER.fine(() -> build + " was not considered trustworthy, but " + revision + " was trusted anyway");
```
daniel-beck, line 984: This seems misleading. Many builds will fall into this category, as their build "causes" will not result in trust, but the revision itself will be trusted.

it seems more natural to me to first check for `#getTrustedRevision` and only check trustworthiness if that is different from the actual `revision`, but I guess it depends on what we consider cheaper to compute?

src/main/java/jenkins/scm/impl/TrustworthyBuilds.java
```diff
@@ -0,0 +1,84 @@
...
+import jenkins.scm.api.TrustworthyBuild;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
+
+public class TrustworthyBuilds {
+
+    // Also effectively handles ReplayCause since that is only ever added in conjunction with UserIdCause. (see ReplayAction.run2)
+    @Extension
+    public static TrustworthyBuild byUserId() {
+        return (build, listener) -> {
+            var cause = build.getCause(Cause.UserIdCause.class);
```
daniel-beck, line 41: While it should be rare, there can be multiple causes of the same type for a build that aren't collapsed into one entry. Whether a build is approved or not depends on the (insertion) order of the `CauseAction#causeBag`.

src/main/java/jenkins/scm/api/SCMSource.java
jglick: `TrustworthyBuild` should be cheaper to compute than `getTrustedRevision`, so I would leave the logic in this order and just clarify the log messages.

src/main/java/jenkins/scm/impl/TrustworthyBuilds.java
jglick: > there can be multiple causes of the same type for a build that aren't collapsed into one entry

The API does not prevent it but this ought never happen—it is the responsible of code triggering the build to pass at most one `Cause` of any given type. At worst a build is not considered trusted when it _could_ have been, so this does not seem like a problem in practice.

daniel-beck (approved):
The behavior introduced here makes sense.

I am not too familiar with the design of SCM API so cannot really comment on whether the code is placed appropriately, but see nothing particularly wrong either.

timja:
> Putting API on hold since there is active design conversation downstream.

Where? It seems over in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/220#issuecomment-1512081371

PR #250, opened 2024-03-12
Add dependency on ASM API plugin

basil:
`SCMHeadMixinEqualityGenerator` depends on ASM, currently from Jenkins core. We'd eventually like to remove ASM from core in favor of the ASM API library plugin, but before we can do that, plugins need to start linking against the ASM API plugin. This PR prepares SCM API by linking against ASM API. As long as core continues to bundle ASM, core's copy will take precedence and this PR will have no effect in production besides forcing users to install the ASM API plugin. But once ASM is removed from core, then the ASM API plugin's copy will become the primary copy of ASM.

### Testing done

`mvn clean verify -Dtest=InjectedTest`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>asm-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>structs</artifactId>
     </dependency>


jglick (approved):
Good for now. Ultimately better to make this unnecessary: https://github.com/jenkinsci/scm-api-plugin/pull/17#discussion_r91125956

basil:
https://github.com/search?ref=simplesearch&type=Code&q=user%3Ajenkinsci+SCMHeadMixIn just shows `git-plugin` so I suppose this could be done by adding a `.equals()` method to `GitBranchSCMHead` and then implementing https://github.com/jenkinsci/scm-api-plugin/pull/17#discussion_r91125956. To allow for a graceful migration I suppose the existing ASM/reflection logic could be retained when a `.equals()` method does not exist, allowing for people to upgrade `git-plugin` over time. When enough have upgraded, the ASM/reflection logic could simply be deleted.

basil:
[JENKINS-72891](https://issues.jenkins.io/browse/JENKINS-72891) FTR

PR #276, opened 2024-07-26
remove dead yui code

mawinter69:
since longer time the repeatable delete is no longer a button created by yui but directly rendered by Jenkins core as button element  So the call to YAHOO.widget.Button.getButton(e.id) will always return undefined. The `layoutUpdateCallback` will make sure that when the delete button is clicked the `traitSectionRule` is called.

### Testing done
Manual testing

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue


jglick (approved):
I was able to configure a multibranch project with this PR and add and delete traits.

PR #282, opened 2024-10-09
Require Java 17 or newer; migrate to EE 9

basil:


pom.xml
```diff
@@ -79,8 +79,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.426.x</artifactId>
-        <version>2839.v003b_4d9d24fd</version>
+        <artifactId>bom-2.462.x</artifactId>
```
jglick, line 82: (pending https://github.com/jenkinsci/bom/pull/3677)

oleg-z:
Pinning jenkins version to 2.479 seems to break plugin installation on earlier LTS versions e.g. 2.462.3

jglick:
@oleg-z right, so do not try to install https://github.com/jenkinsci/scm-api-plugin/releases/tag/698.v8e3b_c788f0a_6 or newer on older versions of Jenkins.

PR #311, opened 2025-03-16
Reduce EE 8 compatibility layer

basil:


jtnord:
@basil is this still work-in-progress?  I am confused as the PR is not draft but has the `work-in-progress` label.  If it is still in progress could you make it draft until it is ready for review?
Thanks

PR #314, opened 2025-03-24
[JENKINS-75069] Add HTML description for HTML display in ObjectMetadataAction class

GGY-AH:
### Testing done
This issue [JENKINS-75069](https://issues.jenkins.io/browse/JENKINS-75069) intially causes the issue [JENKINS-73226](https://issues.jenkins.io/browse/JENKINS-73226) in the Gitea plugin. Its goal is about adding an HTML description method (```getHtmlObjectDescription```) in ```ObjectMetadataAction``` class to provide a proper way to display the content as an HTML one. A new constructor is provided, but the older one remains. The default default behavior also remains (no interface breakdown). It offers the possibility to display the value using safeHTML.  
  
This solution has been discussed through [a refused pull request](https://github.com/jenkinsci/branch-api-plugin/pull/496) with one of the maintainers of Branch API.

### Submitter checklist
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/jenkins/scm/api/metadata/ObjectMetadataAction.java
+++ b/src/main/java/jenkins/scm/api/metadata/ObjectMetadataAction.java
@@ -30,6 +30,7 @@ import hudson.model.InvisibleAction;
 import java.io.Serializable;
 import java.util.Objects;
 
+import jenkins.model.Jenkins;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMNavigator;
 import jenkins.scm.api.SCMRevision;
@@ -53,6 +54,11 @@ import org.kohsuke.stapler.export.ExportedBean;
  *     {@link #getObjectUrl()} to point to the pull request on GitHub, the {@link #getObjectDisplayName()} to provide
  *     the title of the pull request and {@link #getObjectDescription()} to provide the description of the pull request
  *     </li>
+ *     <li>An external {@link SCMSource} implementation that corresponds to a Gitea repository could use the
+ *     {@link #getObjectUrl()} to point to the Gitea repository URL, the {@link #getObjectDisplayName()} to provide
+ *     the repository name, the {@link #getObjectDescription()} to provide the description of repository, and the
+ *     the optional {@link #getHtmlObjectDescription()} to provide the displayed repository description.
+ *     </li>
  * </ul>
  *
  * @since 2.0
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/scm/api/metadata/ObjectMetadataAction.java
```diff
@@ -116,6 +134,21 @@ public String getObjectUrl() {
         return objectUrl;
     }
 
+    /**
+     * Returns the description of the object or {@code null} for the HTML content. Consumers should assume the content
+     * is valid HTML or plain text that needs to be formatted with {@link Jenkins#getMarkupFormatter()}
+     * or escaped with {@link Util#escape(String)} when being included in HTML output.
+     * Note: If this value is {@code null}, Consumers should get the default object description with
+     * {@link #getObjectDescription()}.
```
jtnord, line 142: How will consumer know which is the correct case for the format?  a wrong assumption will likely be a security issue.

The markup formater may not be one for HTML (it could be ascidoc, so if something returns HTML and you run it cia an ADOC processor then you will likely get garbage out

jtnord (changes requested):

src/main/java/jenkins/scm/api/metadata/ObjectMetadataAction.java
GGY-AH: I understand your meaning. Maybe my comment is misunderstandable. What I mean is this field may be whatever you want, but should be used for the HTML display. I use the Jenkins markup formatter in the Branch API to render a nice description field. But maybe I shouldn't say it may be used with the MarkupFormatter ?

jtnord: The use of a markup formatter for things that do not come from a user should be discouraged as things get out of sync (it was a bad historic pattern in some plugins/core that we should not build on).

If this is set by some code system (e.g. prettifying some info from an external system) then in my PoV what sets this should be sanitizing it before it gets here.

Adding @daniel-beck as he has looked at similar elsewhere (and in core) IIRC.

But additionally I am not sure where this is expected to be used.  Taking arbitrary HTML form external systems could completely break the Jenkins UX.

GGY-AH: I understand well better now. The issue appears in the Gitea plugin in the description column of an organization. It authorizes HTML content and you may want to display HTML as is. But for now, it is simply escaped (for security reason) and displays it as plain text. And, for example, when you select the organization, there is a description below, formatted as HTML, when the description in Gitea is written using HTML tags.

Anyway, it is a pleasure to discuss these stuff with you because I have not all the historicals and don't know all the troubles you encountered.

I am open to search for another solution and attend your guidance.

daniel-beck: Using the configured markup formatter is wrong for the reasons mentioned.

In the past I've recommended a dependency on the specific "Safe HTML" formatter as sort of an API handling HTML sanitization, which would sidestep Markdown and similar incompatible formats being configured, but that caused trouble when the rules for the formatter changed in a way that didn't work for the use case.

The only reasonable option seems to be using an HTML sanitizer like https://github.com/OWASP/java-html-sanitizer/ directly.

GGY-AH: OK. If I understand well, I may propose here the use of this formatter in my getHtmlObjectDescription to sanitize the content with authorized elements, attributes... Then this method would return a safe value. Do you agree with that ?

GGY-AH: I think your feedback takes sense, so I will propose a solution based on our discussion and create a new pull request :) It may be refused, but if I don't test, I will never know I guess.

PR #328, opened 2025-05-19
Fix help display behaviors in the trait list UI

dodo920306:
Wrap helpLink and helpArea in a `.help-sibling` container to ensure help content appears under the correct entry, rather than always in the first block. Also moved the helpLink inline with displayName inside the same <td> to improve layout.

- Fixes confusing UX in repeatable entries
- Matches Jenkins' help JavaScript behavior (finds closest `.help-sibling`)

Before:
ezgif-8b255e49809e91

After:
ezgif-8b01f6aaa1c984

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/resources/jenkins/scm/api/form/traits.jelly
+++ b/src/main/resources/jenkins/scm/api/form/traits.jelly
@@ -105,18 +105,18 @@ THE SOFTWARE.
     <d:tag name="body">
       <local:blockWrapper>
         <j:set var="help" value="${descriptor.helpFile}"/>
+        <div class="help-sibling" style="width:100%">
           <local:row>
             <local:td>
-              <div>
-                <b>${descriptor.displayName}</b>
-              </div>
+              <b>${descriptor.displayName}</b>
+              <f:helpLink url="${help}"/>
             </local:td>
-            <f:helpLink url="${help}"/>
           </local:row>
           <!-- TODO: help support is unintuitive; people should be able to see help from drop-down menu -->
           <j:if test="${help!=null}">
             <f:helpArea/>
           </j:if>
+        </div>
 
         <d:invokeBody/>
 
--- a/src/main/resources/jenkins/scm/api/form/traits.jelly
+++ b/src/main/resources/jenkins/scm/api/form/traits.jelly
@@ -63,12 +63,10 @@ THE SOFTWARE.
     <d:tag name="body">
       <div>
         <j:set var="help" value="${descriptor.helpFile}"/>
-        <div class="help-sibling" style="width:100%">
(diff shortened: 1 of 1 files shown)


dodo920306:
@jenkinsci/scm-api-plugin-developers Any update?

src/main/resources/jenkins/scm/api/form/traits.jelly
```diff
@@ -105,18 +105,18 @@ THE SOFTWARE.
...
             <local:td>
-              <div>
-                <b>${descriptor.displayName}</b>
-              </div>
+              <b>${descriptor.displayName}</b>
+              <f:helpLink url="${help}"/>
             </local:td>
-            <f:helpLink url="${help}"/>
           </local:row>
           <!-- TODO: help support is unintuitive; people should be able to see help from drop-down menu -->
```
jglick, line 115: dates to #36 FTR

jglick:
@timja any idea if this `help-sibling` was ever intended as an API?

The whole file seems rather suspicious to me. You should not need all of this boilerplate just to add a list of traits to an object, something that can be done much more easily with the standard `f:repeatableHeteroProperty` as for example in https://github.com/jenkinsci/oic-auth-plugin/pull/644. I think the only thing we get out of this complexity is the visual categorization of traits into sections.

timja (changes requested):
I just checked with master and this was fixed in https://github.com/jenkinsci/scm-api-plugin/pull/338, these changes aren't required.

cc @janfaracik

jglick:
https://github.com/jenkinsci/scm-api-plugin/pull/328#pullrequestreview-3261575094

dodo920306:
@jglick @timja Wait. What? You can't close this. The problem still remains even with the latest version of this plugin!
ezgif-66256d0d19fd86
image

I thought https://github.com/jenkinsci/scm-api-plugin/pull/328#pullrequestreview-3261575094 was talking about the change in the review rather than this whole PR. Anyway, please reopen this 🙏.

dodo920306:
Like this, you mean?

timja:
> Like this, you meant?

Mostly, I fixed the merge for you in https://github.com/jenkinsci/scm-api-plugin/pull/328/commits/a2b65e1cd4081d9823fec2c199818c25e7d4fdf9

timja (approved):
Thanks! I tested on master and reproduced the issue and can no longer reproduce it with this fix.

PR #357, opened 2025-11-13
Stop assigning random `SCMSource.id`s

jglick:
Requires https://github.com/jenkinsci/branch-api-plugin/pull/567 to avoid regressions in multibranch project event dleivery.

compare #19, #37, #41

The design of Jenkins configuration, particularly when configuration-as-code is involved, expects persisted fields to have predictable values. Generating a random value for a field in some nested object to identify it in the future is not reliable. (The most recent example was anomalous behavior in a CloudBees CI HA controller, but there have historically been other problems.) Better would be to actually delete the `id` field and find a different system for dispatching webhook events and other uses, but this looked like a much bigger change that would be difficult to make compatible from both a code and saved settings perspective.


dwnusbaum (approved):

jglick:
I will plan to merge this tomorrow, to make sure I have time to get https://github.com/jenkinsci/branch-api-plugin/pull/567 released as well, since I do not think it is safe for people to update to just this one.

PR #441, opened 2026-02-04
Update parent POM, switch to Renovate, enable dep automerge

jglick:
Ensures that we pick up https://github.com/jenkinsci/plugin-pom/pull/1290 despite https://github.com/dependabot/dependabot-core/issues/14102.

This PR updates the plugin to:
- Use the latest parent POM version (6.2122.v70b_7b_f659d72)
- Switch from Dependabot to Renovate for dependency management, as in https://github.com/jenkinsci/archetypes/pull/886
- (Perhaps) enable GitHub Actions reusable workflows for dependency automerge, as in https://github.com/jenkins-infra/github-reusable-workflows/pull/48

[CloudBees-internal reference](https://cloudbees.atlassian.net/browse/BEE-66540)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alecharp (approved):

.github/renovate.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "github>jenkins-infra/.github:renovate-config"
```
timja, line 4: Don't you need some configuration to open the bom PR as draft so we don't get notified by codeowner review?
see https://github.com/jenkinsci/scm-api-plugin/pull/443

jglick: @timja that would be nice to add to the org-wide config I guess.

timja: Does it need to be a second config in that repo until this is rolled out widely?

jglick: IMHO if a maintainer _wants_ to merge a BOM bump for a specific reason, they should mark it ready for review themselves before approving & merging.

PR #444, opened 2026-05-25
Inline ChangeRequestSCMHead2 interface into ChangeRequestSCMHead

das7pad:
This PR is cleaning up the `ChangeRequestSCMHead`/`ChangeRequestSCMHead2` interface workaround that was needed back before Java 8.

### Testing done

I've reviewed the usages of [`ChangeRequestSCMHead` in the jenkinsci org](https://github.com/search?q=org%3Ajenkinsci%20ChangeRequestSCMHead&type=code) and did not spot any instances that would not satisfy the `ChangeRequestSCMHead2` requirement. So we could remove the default implementation as well.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead2.java
+++ b/src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead2.java
@@ -24,37 +24,12 @@
 
 package jenkins.scm.api.mixin;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
-import jenkins.scm.api.SCMHead;
-import jenkins.scm.api.SCMSource;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 /**
- * Additional attributes of a {@link ChangeRequestSCMHead} that should have been in the original mixin but we are not
- * targeting Java 8 so we cannot add the default methods to the interface and must have an ugly {@code 2} class instead.
+ * Use {@see ChangeRequestSCMHead} instead.
  * @since 2.2.0
  */
-// TODO once Java 8 is baseline move method to ChangeRequestSCMHead with default return value,
-// TODO deprecate this interface and add @Restricted(NoExternalUse.class) (retain empty interface for binary compat)
-public interface ChangeRequestSCMHead2 extends ChangeRequestSCMHead {
-    /**
-     * Returns the {@link ChangeRequestCheckoutStrategy} of this {@link ChangeRequestSCMHead}.
-     *
-     * @return the {@link ChangeRequestCheckoutStrategy}.
-     */
-    @NonNull
-    ChangeRequestCheckoutStrategy getCheckoutStrategy();
-
-    /**
-     * Returns the name of the actual head on the source control system which may or may not be different from
-     * {@link #getName()}. For example in GitHub or Bitbucket this method would return the name of the origin branch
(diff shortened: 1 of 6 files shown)


src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead2.java
```diff
@@ -24,37 +24,12 @@
...
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 /**
- * Additional attributes of a {@link ChangeRequestSCMHead} that should have been in the original mixin but we are not
- * targeting Java 8 so we cannot add the default methods to the interface and must have an ugly {@code 2} class instead.
+ * Use {@see ChangeRequestSCMHead} instead.
  * @since 2.2.0
  */
-// TODO once Java 8 is baseline move method to ChangeRequestSCMHead with default return value,
-// TODO deprecate this interface and add @Restricted(NoExternalUse.class) (retain empty interface for binary compat)
```
jglick, line 37: Should it not be deprecated?

jglick (commented):
There are [usages right in this plugin](https://github.com/search?q=repo%3Ajenkinsci%2Fscm-api-plugin%20ChangeRequestSCMHead2&type=code) that would need to be adapted, namely https://github.com/jenkinsci/scm-api-plugin/blob/24df4bb455f2822fce3513d446facc649f1210e6/src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMRevision.java#L79-L80 and then I think a downstream PR would need to be prepared for https://github.com/jenkinsci/branch-api-plugin/blob/e73a56d397882c63ae06483ba81742c2e39d25d5/src/main/java/jenkins/branch/BranchNameContributor.java#L73-L75

src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMHead2.java
```diff
@@ -24,37 +24,12 @@
...
-     * separate concept of origin name.
-     *
-     * @return the name this {@link ChangeRequestSCMHead} would have if the {@link SCMSource} were configured
-     * against the {@link #getOrigin()} directly and the change request were be discoverable as a regular
-     * {@link SCMHead} or {@link #getName()} if such a concept is not possible in the backing source control system.
-     */
-    @NonNull
-    String getOriginName();
-}
+@Restricted(NoExternalUse.class)
```
timja: Can it be `DoNotUse`? That way it blocks usages in this plugin too?

```suggestion
@Restricted(DoNotUse.class)
```

jglick: `@Restricted` is inappropriate. It is simply `@Deprecated`.

das7pad:
> There are [usages right in this plugin](https://github.com/search?q=repo%3Ajenkinsci%2Fscm-api-plugin%20ChangeRequestSCMHead2&type=code) that would need to be adapted, namely

Thanks for the pointers, that's addressed in d8d6da5. I've also switch the annotation to `@deprecated` in d2ec56f.
 
> and then I think a downstream PR would need to be prepared for https://github.com/jenkinsci/branch-api-plugin/blob/e73a56d397882c63ae06483ba81742c2e39d25d5/src/main/java/jenkins/branch/BranchNameContributor.java#L73-L75

:+1:  I've opened a PR over there:

- https://github.com/jenkinsci/branch-api-plugin/pull/821

